PR #3, opened 2018-02-27
Tagging component

PanSpagetka:
# How Does it look
export 5

screenshot from 2018-05-02 11-42-10

screenshot from 2018-05-02 13-55-44
screenshot from 2018-05-02 13-56-14
screenshot from 2018-05-02 13-56-34
screenshot from 2018-05-02 13-57-00

# Interaction
screencast from 2018-05-02 14-13-25

# How does it work
Tags are made up of two parts, a category and value. Some categories allow for more than one value to be associated. Categories which can have assigned only one value are marked with `*` character.
Tags can be applied to a single item or multiple items at one time depending on the use case.
By selecting Category and Value from select box, you can assign tags to items. You can also remove assigned values by clicking on cross in Tag View. 

# How to use it
1. Create Redux store with  `taggingApp`or Add it to your store.
2. Dispatch action for loading initial state `store.dispatch({ initialState: defaultState, type: "UI-COMPONENTS_TAGGING_LOAD_STATE" });`
3. Render Component, you can use either  ` TaggingConnected` or `TaggingWithButtonsConnected`, those are the components which work with Redux and contains logic(they are working).

```
  ReactDOM.render(
    <Provider store={store}>
      <TaggingWithButtonsConnected />
    </Provider>,
    document.getElementById('demo-app'),
  );
```
initialState must be structured like this:
```
tags : [
  { description: 'Number', id: 2, singleValue: true,
    values: [{ description: '1', id: 21 }, { description: '2', id: 22 }] },
  { description: 'Animal', id: 3,
    values: [
      { description: 'Duck', id: 31 }, { description: 'Cat', id: 32 }, { description: 'Dog', id: 33 }
    ] 
  },
], 
assignedTags:  [
  { 
    tagCategory: { description: 'Animal', id: 3 },
    tagValues: [{ description: 'Cat', id: 32 }] 
  }
],
affectedItems : [123456, 654321]
```
Name | Type | Default | Mandatory | Description
------- | ------- | ------- | ------- | ------- 
tags | Array | [] | Yes |  Categories and its Values which can be assigned to items
description | String | None | Yes | Text description of each Category, this text be showed in Select.
id | Int(but it should work with others) | None | Yes | Unique id used to distinguish Categories between themself. 
singleValue | Bool | false | No | Only single value for this category can be assigned at one time. These are marked with `*` in UI.
values | Array | None | Yes | Values which can be assigned to this particular Category. Contains string `description` and Int `id`
assignedTags | Array | [] | No | Categories  and it's Values, which are already assigned to `affectedItems`
affectedItems | Array | [] | No | Array of ids identifying items, which tags are edited


.eslintrc
```diff
@@ -12,6 +12,9 @@
       "ignoreUrls": true,
       "ignoreComments": false
     }],
+    "react/require-default-props": [0],
```
ohadlevy: while i dont have an opinion, maybe linting rule changes should happen in separate pr's so their impact could be discussed?

src/tagging/components/tag.jsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+ import style from './style';
```
ohadlevy: nitpick: extra whitespace.

src/tagging/components/tag.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+class Tag extends React.Component {
```
ohadlevy: imho you dont need a class here, you can simply use a const function instead?

src/tagging/components/tag.jsx
```diff
@@ -0,0 +1,29 @@
...
+  handleClick = () => this.props.onTagDeleteClick(this.props.tagCategory, this.props.tagValue);
+
+  render() {
+    return (
+      <li key={`${this.props.tagCategory}: ${this.props.tagValue}`}>
+        <span className="label label-info">
+          {this.props.tagCategory}: {this.props.tagValue}
+          <a href="#" onClick={this.handleClick}>
+            <span className="pficon pficon-close" aria-hidden="true" />
```
ohadlevy: please use the icon component from patternfly-react

src/tagging/components/tag.jsx
```diff
@@ -0,0 +1,29 @@
...
+            <span className="sr-only">Remove</span>
```
ohadlevy: all strings should be available as props, so later on  you can i18n them.

src/tagging/components/tagModifier.jsx
```diff
@@ -0,0 +1,42 @@
...
+import PropTypes from 'prop-types';
+import TagSelector from './tagSelector';
+import ValueSelector from './valueSelector';
+
+const TagModifier = (props) => {
+  const tagValues = props.tags[props.selectedTagCategory] || [];
+  const tagCategories = Object.keys(props.tags) || [];
+  return (
+    <div className="col-lg-6">
+      <div className="row">
```
ohadlevy: please use col and row components from patternfly-react

src/tagging/components/tagModifier.jsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import TagSelector from './tagSelector';
+import ValueSelector from './valueSelector';
+
+const TagModifier = (props) => {
```
ohadlevy: please extact the strings, so you can do 
```js
const TagModifier = ({ tags, selectedTagCategory, onTagCategroyChange... })
```
and then just use them without the props.<>

src/tagging/reducers/index.js
```diff
@@ -0,0 +1,19 @@
+import { combineReducers } from 'redux';
+import { modifySetTags, toggle } from './reducers';
+
+const TAGS = {
```
ohadlevy: these should move into fixtures files.

src/tagging/reducers/reducers.js
```diff
@@ -0,0 +1,24 @@
+// state = state.setTags
+export const modifySetTags = (state = [], action) => {
```
ohadlevy: imho you should use immutable library to avoid changing the existing state.

PanSpagetka:
@ohadlevy I think this could be part of patternfly react in the future, but not yet. First I would like to test it in ManageIQ and see if there are some changes to be made.

After I test it, I would move it to patternfly, preferably to monorepo(when it's done). Because I think this component have very specific usecase and it is not necessary for everybody who uses patternfly to load it.

PanSpagetka:
I have probably should have marked it as WIP, because there is still some work to be done.

@miq-bot add_label WIP

src/tagging/components/tag.jsx
PanSpagetka: I will use https://github.com/patternfly/patternfly-react/pull/223 as soon as it's merged

.eslintrc
PanSpagetka: https://github.com/ManageIQ/react-ui-components/pull/6

src/tagging/actions/actions.js
```diff
@@ -0,0 +1,5 @@
+// action creators
+export const TOGGLE_TAG_CATEGORY_CHANGE = 'TOGGLE_TAG_CATEGORY_CHANGE';
```
karelhala: Please add prefix for these actions so we don't clash with other action types.
```javascript
export const TOGGLE_TAG_CATEGORY_CHANGE = 'UI-COMPONENTS_TOGGLE_TAG_CATEGORY_CHANGE';
export const TOGGLE_TAG_VALUE_CHANGE = 'UI-COMPONENTS_TOGGLE_TAG_VALUE_CHANGE';
export const DELETE_SET_TAG = 'UI-COMPONENTS_DELETE_SET_TAG';
export const ADD_SET_TAG = 'UI-COMPONENTS_ADD_SET_TAG';
```

src/tagging/reducers/index.js
PanSpagetka: This shouldn't be part of this component, It was there because of testing purposes. I moved it to stories and demo. Now I pass it as a default state, when creating this component

src/tagging/components/tag.jsx
PanSpagetka: Done

src/tagging/components/tag.jsx
PanSpagetka: deleted

src/tagging/components/tag.jsx
PanSpagetka: Done, strings are available as props with default values.

src/tagging/components/tagModifier.jsx
PanSpagetka: Done

src/tagging/components/tagModifier.jsx
PanSpagetka: Done

serenamarie125:
@ohadlevy @PanSpagetka I think someone already implemented something in the managieq-ui-service repo in conjunction with UX from a design perspective.  @AllenBW I think there was a tag editor modal submitted - I couldn't find a PR to reference.  Might you have it handy?

serenamarie125:
@ohadlevy re: your PF React suggestion, I agree that this would be a great contribution!

AllenBW:
@serenamarie125 yeah!! that was a while back... https://github.com/ManageIQ/manageiq-ui-service/pull/402

serenamarie125:
Actually, [here's](https://github.com/ManageIQ/manageiq-ui-service/pull/402) the initial piece of it ... although it looks like it never went through a ux review
@beanh66 this is relatively aligned with [the design](https://manageiq.github.io/manageiq-design/UX/common/Tags/design) right?

PanSpagetka:
@serenamarie125 Thanks for showing me this, I didn't know that we have official design. But I still think my PR is needed. Even though my knowledge of angular is very poor, I don't think, that somebody can take that Tag editor from ServiceUI and use it in ManageIQ-ui-classic(or anywhere else) easily. It would require some amount of work. 

The point of my PR is make reusable component, composed of several simpler components creating some logical structure, using preferable technologies(React, Redux). So you can use this component, if I oversimplify it, with one line of code.

Because logic, data and design are separated, you can change design of the component without affecting it's functionality easily. Or you can simply add another component into it. For example View of Affected Entities, without need for modifying existing code, only adding new.

As I said earlier, I would like to(and will) make PR into Patternfly React, when I have working example in ManageIQ.

If you have some suggestions about design, please let me know.
I hope I have explained why I was(am) making this PR :D

beanh66:
@serenamarie125 correct, that MIQ implementation is in line with the design we had. Thanks!

serenamarie125:
@PanSpagetka 
Is the screenshot here from your implementation?

Could we get closer to the design shown here https://manageiq.github.io/manageiq-design/UX/common/Tags/design?

Is it possible to have multiple values assigned to the same category?  In the design, we are showing a design where we would show that in the "chip".

PanSpagetka:
@serenamarie125 Yes it is. And I will make it look more like in the design. @terezanovotna offered me help, so when I am done with coding I will redesigned it according to the https://manageiq.github.io/manageiq-design/UX/common/Tags/design

src/tagging/components/taggingWithButtons.jsx
```diff
@@ -0,0 +1,99 @@
...
+          assignedTags={this.props.assignedTags}
+          onTagValueChange={this.props.onTagValueChange}
+          onTagMultiValueChange={this.props.onTagMultiValueChange}
+          onTagCategoryChange={this.props.onTagCategoryChange}
+          onTagDeleteClick={this.props.onTagDeleteClick}
+          selectedTagCategory={this.props.selectedTagCategory}
+          selectedTagValue={this.props.selectedTagValue}
+        />
+        <Row className={'pull-right'}>
+          <ButtonGroup>
```
karelhala: I would put ButtonGroup in seperate component so we have easier way of combining these two components together.

src/tagging/components/valueSelector.jsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import Select from 'react-select';
+import PropTypes from 'prop-types';
+
+
+class ValueSelector extends React.Component {
+  handleChange = (selectedOption) => {
```
karelhala: Refactor this component to be simpler
```javascript
handleChange = (selectedOption) => {
    this.props.onTagValueChange({ description: selectedOption.label, id: selectedOption.value });
  };

  value = this.props.selectedOption.id;
  label = this.props.selectedOption.description;
  tagValues = this.props.tagValues.map(tag => ({ value: tag.id, label: tag.description }));

  render() {
    return (
      <Select
        name="form-field-name"
        value={this.value}
        label={this.label}
        onChange={this.handleChange}
        options={this.tagValues}
        clearable={false}
      />
    );
  }
```

src/tagging/reducers/reducers.js
```diff
@@ -0,0 +1,54 @@
+import * as actionsConstants from '../actions/actions';
+// state = state.assignedTags
+export const modifyassignedTags = (state = [], action) => {
+  switch (action.type) {
+    case actionsConstants.DELETE_ASSIGNED_TAG:
+      return [...state.filter(tag => (tag.tagCategory.id !== action.tag.tagCategory.id)),
```
karelhala: Please refactor this to use functions and make it simpler.

src/tagging/components/tagCategory.scss
```diff
@@ -0,0 +1,12 @@
+ul.tag-category.list-inline {
+  display: inline-block;
+  background-color: #00659c;
+  padding: 5px 5px 5px 5px !important;
```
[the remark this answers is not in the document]
PanSpagetka: Yes, it is, because of use in ManageIQ

demo/demo-app.jsx
```diff
@@ -1,7 +1,35 @@
...
+  },
+];
+
+const newAssignedTags = [{ tagCategory: { description: 'Name', id: 1 }, tagValues: [{ description: 'Pepa', id: 11 }] }];
+
+const defaultState = { tags: newTags, assignedTags: newAssignedTags };
 
 export default function renderApp() {
-  ReactDOM.render(<cmpA.HelloCmp />, document.getElementById('demo-app'));
+  const store = createStore(taggingApp);
```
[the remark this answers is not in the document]
karelhala: Resolved

PanSpagetka:
@remove_label wip

PanSpagetka:
@miq_bot remove_label wip

serenamarie125:
@PanSpagetka can you add a couple add a storybook so that we can access this to test?  @jeff-phillips-18 this should have a designer assigned to it for review, since this component doesn't exist in PF.  Add me if you'dlike

karelhala:
@serenamarie125 screenshots sure could help, or even gifs. However we do not have continous delivery set up, because there is no particular tool for it. Abusing github pages is not good, it might be okay for master to show how upstream is working, but for PRs it's not great. There are ways how to solve it and abuse github pages for it, but that's not good and they shouldn't be used in this way. However if you wish to see how such component is working you can pull this PR and run storybook locally, if you want we can give you either list of commands which to run to see it locally or we could provide you with easy script which would do it in one step.

ohadlevy:
how about using rawgit with storybook? the same tools we use in patternfly-react?

karelhala:
@ohadlevy that's somewhat OK, but it has a lot of drawbacks.

* Developer has to run `npm run build-storybook` everytime he makes changes - if this would be the only problem I could live with that
* We have to expose storybook changes to github (biggest problem)
* Storybook changes will be part of code to review (higher number of changes per PR)
* Github is against it - https://github.com/rgrove/rawgit/blob/master/FAQ.md#why-is-this-necessary-cant-i-just-load-files-from-github-directly

To be honest I can't find any automated thing which would serve static content both for upstream and for PRs. I know that many projects are using githu pages and rawgit for these purposes, but it's lacking some automation. Developer has to run certain commands and has to push files which should not be pushed. However this is not part of this PR, so I'll file an issue where we could discuss this. I believe @himdel and @priley86 will be interested in this as well.

PanSpagetka:
@serenamarie125 I have added more screenshots and also gif so you can see interaction. Colors in the gif are for some reason different, you can see real colors in the screenshots.

ohadlevy:
@karelhala I'm not sure i understand your point, storybook can be configured to push to rawgit (which is not a github service afaiu). further, you can have a unique urls pr PR/branch, so they dont conflict with each other.

if its an isolated component, it actually makes sense to develop it inside storybook first (much faster), then you can see the visual interaction based on mocked data. a second step could be to connect it to the API etc.
TBH - i see storybook as a useful tool, and this is used often in both patternfly-react and foreman repos.

src/tagging/components/tagging.jsx
```diff
@@ -0,0 +1,87 @@
...
+import { Grid, Row, Col } from 'patternfly-react';
+import TagModifier from './tagModifier';
+import TagView from './tagView';
+import CategoryModifier from './categoryModifier';
+import ValueModifier from './valueModifier';
+
+class Tagging extends React.Component {
+  onTagValueChange = (selectedTagValue) => {
+    const action = { tagCategory: this.props.selectedTagCategory, tagValue: selectedTagValue };
+    if (this.findSelectedCat(this.props.selectedTagCategory).singleValue) {
```
serenamarie125: When limited to single select, we are replacing the value, correct?  Just want to make sure we aren't forcing the user to delete previous value before adding a new one

src/tagging/tests/__snapshots__/tagging.test.js.snap
```diff
@@ -0,0 +1,335 @@
...
+    </div>
+  </div>
+  <div
+    className="row"
+  >
+    <div
+      className="col-md-12"
+    >
+      <div>
+        * Only a single value can be assigned from these categories
```
serenamarie125: could we have an information icon in the dropdown rather than an '*' ?  If so, we could provide hover text status that it's single select instead of having it on the dialog.  What do you think?

src/tagging/components/valueModifier.jsx
```diff
@@ -0,0 +1,35 @@
...
+ValueModifier.propTypes = {
+  selectedTagValue: PropTypes.object.isRequired,
+  onTagValueChange: PropTypes.func.isRequired,
+  valueLabel: PropTypes.string,
+  multiValue: PropTypes.bool,
+  tagValues: PropTypes.arrayOf(PropTypes.object),
+};
+
+ValueModifier.defaultProps = {
+  valueLabel: 'Assigned Value:',
```
serenamarie125: I think this can be "Value" (drop assigned).  Also I'd remove the colon if this is for future product work - if it's for current ManageIQ I'd see keeping it for consistency.  If you remove it, make sure you remove it from Category as well.

serenamarie125:
@PanSpagetka here's some input from our visual designer.  Might you be able to make these tweaks as well?  

The PF button sizes are included alongside the tags, as a size reference. It might make the most sense to treat the tag container with "default button" sizing, and then the individual tags with "xs button" sizing. (This is so that the tags don't seem way to large and clunky alongside or underneath a row of buttons. Everything would have the same height!) I've called out the background colors, as well.  "pf-blue-500" for the tag container, and "pf-blue-400" for individual tags (which actually teeters the line of compliance/non-compliance in a contrast test!).

tagsmorespacing

tagsspacing

There's also a few spacing tweaks shown in the image.  Thank you!

src/tagging/components/tagging.jsx
PanSpagetka: Yes, it just replace the value. User doesn't have to delete previous value.

src/tagging/tests/__snapshots__/tagging.test.js.snap
PanSpagetka: Yep, good idea.

src/tagging/components/valueModifier.jsx
PanSpagetka: Done.

PanSpagetka:
@serenamarie125 Thank you for review.

1. Done
2. Done
3. Yes, buttons are optional part of the pattern. You can have exactly same component with or without buttons. First screenshot, with colorful rectangles, is older. I have adjusted button order to be consistent ManageIQ. But the spacing is problematic, I am using button group component which doesn't support spacing. I will look at it later.
4. Yep, there is.

I have addressed the visual feedback, changed spacing, colors, alignments  and updated screenshots/gif.

Issue #2, opened 2018-02-27
components should not be exposed via window


ohadlevy:
imho, we should not continue using window as a way to share code, but rather use webpack import to handle it for us. (https://github.com/ManageIQ/react-ui-components/blob/master/src/vendor.js#L12)

the main motivation for me is to find errors during build time vs run time.

karelhala:
Whenever we see component accessing window object we can't merge such PR, however in vendor.js we store some libraries into window object so we can access them via import like this `import * as React from 'react'` and when we use [externals/#object](https://webpack.js.org/configuration/externals/#object) final build will not include these libraries.

I couldn't find better sollution to access these libraries exposed with externals and they ended up as part of our build, which is not acceptable.

ohadlevy:
I'm not sure i follow, which error did you face? you can always use webpack to build and reference via alias the path to find those libraries.

btw, having said that, since this repo should be consumed as a package, imho it should not import its own redux/react etc as it would most likely conflict with the packages the core repo requires.

karelhala:
> it should not import its own redux/react etc as it would most likely conflict with the packages the core repo requires

That's exactly the issue I was facing. We have to import react/redux/lodash/whatever so we can check errors during build time, but we can't include them into final build but import these packages from application which imports this library. That's what https://webpack.js.org/configuration/externals/ is for. However some libraries were not behaving correctly and I could not access them, webpack tried to outsmart me and included them as part of build (which wasn't correct), so I stored these packages into window object and configured webpack in a way that he will load these packages from window.

So long story short
 * we can't import other libraries into final build - use externals.
 * some libraries can't be loaded correctly using externals - store these libraries into window object.

ohadlevy:
TBH I disagree, that's why you have peer dependencies in your package.json file.

karelhala:
You are right, that we should add `react` and other libraries to peerDependencies, but still when running build with webpack these dependendencies would be bunled with our library, which is no go. We also don't want to watch for these external libraries because build would be too slow.

So the problem is not to include external libraries into final build, but we want to use imports and not access window properties.

I tried couple of approaches
 * Using https://webpack.js.org/plugins/commons-chunk-plugin/ - won't work because in final build webpack would try to load them, but they'll be missing.
 * Using separate build process for vendor's libraries - using ES6's `import` is not working.
 * Loading external libraries from CDN - using ES6's `import` is not working.

[...]

PR #9, opened 2018-03-05
Remove jquery and lodash from vendor dependencies

karelhala:
### Link to https://github.com/ManageIQ/react-ui-components/issues/2
As addressed in issue #2 we shouldn't include so many external libraries into our package, we don't need (right now) lodash and jQuery. So remove them from dependencies and from vendor file.

### UI changes
none


package.json
```diff
@@ -89,12 +89,10 @@
     "webpack-dev-server": "^2.11.1"
   },
   "dependencies": {
-    "lodash": "^4.17.4",
     "patternfly-react": "^1.8.0",
     "react": "^16.2.0",
     "react-dom": "^16.2.0",
     "prop-types": "^15.6.0",
-    "react-props": "0.0.3",
     "react-select": "^1.2.1",
```
ohadlevy: should remove jquery here too.

karelhala: Good point, forgot to include it. :+1:

PR #13, opened 2018-04-26
Component: vm snapshot add form component

Hyperkid123:
Created component using react-final-form library.

Component should replace old ruby form in ui-classic. 

compute -> infrastructure -> virtual machines
select virtual machine, click on snapshot row in table
click on add snapshot button in top toolbar

### DIFF
--- a/src/vm-snapshot-form/finalFormTextArea.jsx
+++ b/src/vm-snapshot-form/finalFormTextArea.jsx
@@ -1,0 +1,28 @@
+import React from 'react';
+import { FormGroup, HelpBlock, Col, FormControl } from 'patternfly-react';
+import PropTypes from 'prop-types';
+import { inputObjectProps, metaObjectProps } from './finalFormPropTypes';
+
+const FinalFormTextArea = ({
+  meta,
+  input,
+  label,
+}) => (
+  <FormGroup validationState={meta.error ? 'error' : null}>
+    <Col xs={2} componentClass="label" className="control-label">
+      {label}
+    </Col>
+    <Col xs={8}>
+      <FormControl componentClass="textarea" {...input} />
+      {meta.error && <HelpBlock>{meta.error}</HelpBlock>}
+    </Col>
+  </FormGroup>
+);
+
+FinalFormTextArea.propTypes = {
+  input: inputObjectProps,
+  meta: metaObjectProps,
+  label: PropTypes.string.isRequired,
+};
+
+export default FinalFormTextArea;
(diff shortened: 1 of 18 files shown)


Hyperkid123:
@miq-bot assign @karelhala

Hyperkid123:
@karelhala should i move the generic form inputs to some other folder? Should i add them to dist?

src/vm-snapshot-form/finalFormTextArea.jsx
```diff
@@ -0,0 +1,28 @@
...
+import { FormGroup, HelpBlock, Col, FormControl } from 'patternfly-react';
+import PropTypes from 'prop-types';
+import { inputObjectProps, metaObjectProps } from './finalFormPropTypes';
+
+const FinalFormTextArea = ({
+  meta,
+  input,
+  label,
+}) => (
+  <FormGroup validationState={meta.error ? 'error' : null}>
```
ohadlevy: DRY? maybe something like https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/react_app/components/common/forms/CommonForm.js ?

Hyperkid123: This is actually something that i am wondering about myself. Current behavior is consistent with forms in manageiq-ui-classic, but i would also prefer something similar. I think we should talk about this more, and decide if we wan't different behavior in new forms.

ohadlevy:
@karelhala I don't understand your motivation? which consistency and why does it matter to keep it like that? (don't forget that other users such as v2v are using this approach).

overall, i think its false to start a new solution based on wrong foundations.... more people will copy paste this solution. we are not under pressure - so why not do it right?

karelhala:
@ohadlevy yes you are right but this PR addresses this https://bugzilla.redhat.com/show_bug.cgi?id=1527371 where it's stated that when page loads it's missing these validations. We need to keep it this way and convince others that we should add touched as well. Which will eventually be easy and we'd fix this in all forms.

skateman:
We have *N* forms that use the *"bad pattern"* and a single one that uses the *"good pattern"*, so with this change we got *N+1 vs 1* where *N* is a really huge number. I agree with @ohadlevy that we should fix this, but increasing the number on the right side of the equation by one just introduces more inconsistency, both visually and codewise. So what about making all the forms first easy to maintain and then (╯°□°）╯︵ ┻━┻ everything to the right side as a separate task. Also we should communicate this with QE as they test against the *"bad pattern"*.

PR #14, opened 2018-05-03
React components for ManageIQ textual summary

martinpovolny:
This is a continuation of https://github.com/ManageIQ/manageiq-ui-classic/pull/3420

  * I am moving the components to the components repo, trying to decouple the specifics.
  * Adding more specs,
  * Adding simple storybook


package.json
```diff
@@ -90,7 +90,7 @@
   },
   "dependencies": {
     "lodash": "^4.17.4",
-    "patternfly-react": "^1.8.0",
+    "patternfly-react": "^1.16.4",
```
ohadlevy: might make sense to bump to 2.x as it already introduced api changes (for something we dont use yet - but can't see the value of using an older version now).

src/textual_summary/data/operation_ranges.js
```diff
@@ -0,0 +1,15 @@
+export const operationRangesData = {
+  title: 'Operation Ranges',
+  items: [
+    {
+      label: 'CPU',
```
ohadlevy: should these strings be extracted? or are they used as props somewhere else?

src/textual_summary/empty_group.jsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+
+export default function EmptyGroup(props) {
+  return (
+    <table className="table table-bordered table-striped table-summary-screen">
```
ohadlevy: How about using patternfly tables? https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?selectedKind=Table&selectedStory=PatternFly%20Table%20Styles&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

src/textual_summary/data/operation_ranges.js
martinpovolny: This is test data. It  is used in stories and tests.

src/textual_summary/empty_group.jsx
martinpovolny: I can do it, but not in the first implementation. My goal in this PR is to create React components from HAML files.

package.json
martinpovolny: I need this compatible with ui-classic. The goal is to replace textualsummaries in ui-classic.

ohadlevy:
my main concern is that it things that are kept for next version are often
skipped, also even if you are not using the pf table implementation, I do
think you need a table component instead of using html.

martinpovolny:
1. I don't think that this is a problem I have. 
  1. I have other things that I need to do now.
  1. Size (scope) of a PR is to be limited. A step at a time. I am already splitting the original PR into 4 parts. I don't feel I should be adding more to the scope now.

src/textual_summary/textual_group.jsx
```diff
@@ -0,0 +1,74 @@
...
+import PropTypes from 'prop-types';
+import GenericGroup from './generic_group';
+import TagGroup from './tag_group';
+import SimpleTable from './simple_table';
+import OperationRanges from './operation_ranges';
+import MultilinkTable from './multilink_table';
+import TableListView from './table_list_view';
+import EmptyGroup from './empty_group';
+
+const renderComponent = (props) => {
```
karelhala: Maybe rewriting this to hash might be better than using switch
```jsx
const componentMapper = {
  GenericGroup: props => (
    <GenericGroup onClick={props.onClick} items={props.group.items} title={props.group.title} />
  ),
  TagGroup: props => <TagGroup items={props.group.items} title={props.group.title} />,
  SimpleTable: props => (
    <SimpleTable
      labels={props.group.labels}
      rows={props.group.rows}
      title={props.group.title}
    />
  ),
  OperationRanges: props => <OperationRanges items={props.group.items} title={props.group.title} />,
  MultilinkTable: props => (
    <MultilinkTable
      onClick={props.onClick}
      items={props.group.items}
      title={props.group.title}
    />
  ),
  TableListView: props => (
    <TableListView
      rowLabel={props.group.rowLabel}
      onClick={props.onClick}
      title={props.group.title}
      headers={props.group.headers}
      values={props.group.values}
      colOrder={props.group.colOrder}
    />
  ),
  EmptyGroup: props => <EmptyGroup title={props.group.title} text={props.group.text} />
};
```
That way you can specify which component is using which propTypes
```javascript
componentMapper.GenericGroup.propTypes = {
  group: PropTypes.shape({
    title: PropTypes.string.isRequired,
    items: PropTypes.any,
  }).isRequired,
  onClick: PropTypes.func.isRequired,
};

componentMapper.TagGroup.propTypes = {
  group: PropTypes.shape({
    title: PropTypes.string.isRequired,
    items: PropTypes.any,
  }).isRequired,
};
```
Or use [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) and inherit these properties.

martinpovolny: > Maybe rewriting this to hash might be better than using switch

If you mean a simple hash, that it would not work, because a hash is not lazy evaluated. All the branches would get called and then the apropriate result would be selected. So that approach would crash on the propTypes mismatch (different case branches need diffferent properties).

martinpovolny: Oh, you mean a hash of functions. Then yes, it woudl work but is a hash of functions actually better than what I have now?

martinpovolny: Anyway. If it's a 'maybe changing it will look better' then I am not changing it unless you insist. I see the difference as marginal.

PR #20, opened 2018-05-10
Created common form fields render functions.

Hyperkid123:
Moved form fields render function to common location. This will provide generic components for the forms based on Final Form library.


karelhala:
Released in v0.3.0

PR #21, opened 2018-05-10
Added validate on mount option to form fields.

Hyperkid123:
Added prop to form fields. This prop specifies if, the field should be validated onMount, or after it is touched.
Currently waiting for #20


Hyperkid123:
@miq-bot remove_label unmergeable

Hyperkid123:
@miq-bot assign @karelhala

PR #22, opened 2018-05-15
Added select component for final form.

Hyperkid123:
This is a wrapper around the `react-select` component, customizing it for the `FinalForm` requirements. So far it handles only single-select. Will need some styling optimization to properly support multi select.


Hyperkid123:
@miq-bot assign @karelhala

PR #25, opened 2018-05-21
TextualSummary: more stories, tests and exports.

martinpovolny:
The primary goal of this PR is to export more components that need to be used one-by-one in the ManageIQ.

Secondary, more tests and storybooks are provided.


src/textual_summary/generic_table_row.jsx
```diff
@@ -47,7 +47,9 @@ GenericTableRow.propTypes = {
     title: PropTypes.string,
     image: PropTypes.string,
     icon: PropTypes.string,
-    value: PropTypes.any,
+    label: PropTypes.any,
```
Hyperkid123: We should avoid using `PropTypes.any`. I know that sometimes it can be difficult, especially dorm some dynamic objects, but props like `label` should be defined properly.

Also, if you don't mark object attributes as required, prop validation will pass even for empty object. If there are some conditions, when should be attribute required, and whet it should be undefined, you can create custom prop validator. You can  check this example: https://reactjs.org/docs/typechecking-with-proptypes.html

src/textual_summary/stories/GenericGroup.stories.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import GenericGroup from '../generic_group';
+import { genericGroupData } from '../data/generic_group';
+
+storiesOf('TextualSummary', module)
+  .add('GenericGroup', () => {
+    return (
+      <GenericGroup items={genericGroupData.items} title={genericGroupData.title} onClick={e => null} />
```
Hyperkid123: What is the purpose of `onClick={e => null}`? I would advise specifying this inside the `GenericGroup` component and simply don't pass any props.

You can use defaultProps of that component. Something like
```
GenericGroup.propTypes = {
  onClick: PropTypes.func
...
}

GenericGroup.defaultProps = {
  onClick: () => {}
}
```

src/textual_summary/stories/MultilinkTable.stories.js
```diff
@@ -0,0 +1,16 @@
...
+import MultilinkTable from '../multilink_table';
+import { multilinkTableData } from '../data/multilink_table';
+
+storiesOf('TextualSummary', module)
+  .add('MultilinkTable', () => {
+    return (
+      <MultilinkTable
+        title={multilinkTableData.title}
+        items={multilinkTableData.items}
+        onClick={e => null}
```
Hyperkid123: again `e => null`, not a fan

src/textual_summary/stories/GenericGroup.stories.js
martinpovolny: For a real use the `onClick` is mandatory. In the Storybook, I don't care.

I can replace it with some `alert("You clicked!")` if you want. But it looks silly.

No I don't want to have a default value. For a real use there needs to be a handler.

See https://github.com/ManageIQ/manageiq-ui-classic/pull/3420/files#diff-ea54a250fdb683370b39733f087cfbdeR5

There I have a default on-click handler. BUT: that behavior is manageiq-ui-classic specific internal. For that reason it's there and not here.

Does that address your concerns?

src/textual_summary/stories/MultilinkTable.stories.js
martinpovolny: The explanation remains the same here ;-)

src/textual_summary/generic_table_row.jsx
martinpovolny: Right. I did not implement describe all the data types in a lot of detail. I plan to improve that in follow up PRs. Together with the use of patternfly-react components where applicable.

src/textual_summary/stories/GenericGroup.stories.js
Hyperkid123: Ahh ok i thought that there is some rule. Never mind then

Hyperkid123:
@karelhala can you merge this?

src/textual_summary/stories/GenericGroup.stories.js
karelhala: I don't like `e => null` as well, since it does not give any information to user of such storybook. Can you please use https://github.com/storybooks/storybook/tree/master/addons/actions so the action is logged at least into storybook action list?

himdel:
Uh, sorry, timing :) so .. what Karel [said](https://github.com/ManageIQ/react-ui-components/pull/25#discussion_r190208890)

karelhala:
Btw @martinpovolny if you want to show this storybook to outsiders, you can do it by using http://netlify.com from your repository and with build command as `npm run vendor; npm run build-storybook`. Once the first build starts to run you can add deploy to all branches you push to your repository.

PR #29, opened 2018-05-24
Add new user form for Rbac

Hyperkid123:
Created a react implementation of old ruby form. Original form is at ops/access control/users/add new user.

- [x] i18n mock #32 
- [ ] replace test renderer #39 

storybook: https://5b1f8eacc9659226588a0774--confident-banach-ee0a7a.netlify.com/?selectedKind=Rbac%20forms&selectedStory=Rbac%20add%20user%20form&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel


Hyperkid123:
@karelhala lots of snapshot changes are here because the automatic id adding. Only adds `id` tah to element. Need ids for input form testing.

martinpovolny:
I see a number of strings w/o i18n.

src/forms/finalFormComponent.jsx
```diff
@@ -100,17 +100,39 @@ const inputTypeProp = (props, propName) => {
   return undefined;
 };
 
+const optionsPropType = (props, propName, componentName) => {
+  if (props.componentType === 'select') {
+    if (!Array.isArray(props.options)) {
```
martinpovolny: all this custom validation code is needed because it applies only to `componentType === 'select'` ?

Hyperkid123: i plan to expand it. In select component, you can specify the name of the key and value attribute of options. A put there in preparation in the future. I wanted to include it into this PR, but for some reason it wasn't working properly. Should i remove it?

martinpovolny: No need to remove it. I am just trying to understand it ;-)

martinpovolny:
Can you, please, rebase? Let's get this merged.

Hyperkid123:
@martinpovolny just a small info. I have been talking to @terezanovotna and she agreed that creating/editing users is too complicated, so we will try to figure out some better way to manage that.

martinpovolny:
Well I thought we where trying to validate that the final form stuff is useful to implement what we already have in the product.

What you are telling me is that you are redesigning the creating/editing of users. That's a different scope in my oppinion.

Hyperkid123:
@martinpovolny you're right, but i will have to revisit this later then.

Hyperkid123:
@martinpovolny so the UI part is finished. In order to use this inside manageIQ, we have create proper API endpoints for user groups and saving.

This cannot be used for editing just yet. Because of some special features of the current editing form. Currently, the edit form is used not only for adding/editing, but also for displaying information about user, some additional Smart Management information and redirecting to group/role information pages. Not saying that it is difficult to implement, but i think it should be split into multiple independent components.

terezanovotna:
@Hyperkid123 do we have a storybook so I can see it?

Hyperkid123:
@terezanovotna i have also forgot, that online storybooks are broken at the moment. So a screenshot is the best i can do for now...
vystrizek

Or you can checkout this pr and run the storybook locally

terezanovotna:
@Hyperkid123 when you get the storybook working, let me know! also, i will be in the office on Tuesday!

I have one thing: when adding an email, can we not make a yellow background and have it similar to [this?](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/form.html). When you type the text, the border is slighly blue. and when you click on another field, it turns back to grey

Hyperkid123:
@terezanovotna oh thats not yellow that's just google chrome autocomplete. And the inputs are the same as in patternfly. in this repo it is just a wrapper around patternfly components to make it compatible with the form API.

terezanovotna:
@Hyperkid123 hah, that is a good explanation! great job on this!

Hyperkid123:
@terezanovotna storybook is in first comment.

terezanovotna:
Created a design issue about this component and it looks like it's a duplicate. Interested in seeing more about it because I like this implementation @Hyperkid123 

https://github.com/patternfly/patternfly-design/issues/679

PR #30, opened 2018-05-25
Textual Summary fixes and improvements

martinpovolny:
Number of textual summary fixes and improvements:

 * fix rows and cols (thanks to @skateman for testing)
 * convert values (that might be bools) to string (thx @skateman)
 * use storybook/action (feedback from @karelhala in https://github.com/ManageIQ/react-ui-components/pull/25#discussion_r190208890)

### DIFF
--- a/src/textual_summary/generic_table_row.jsx
+++ b/src/textual_summary/generic_table_row.jsx
@@ -11,7 +11,7 @@ const renderMultivalue = function renderMultivalue(values, onClick) {
             <tr onClick={e => onClick(item, e)} key={i} className={item.link ? '' : 'no-hover'} title={item.title}>
               <td style={{ border: 0, margin: 0, padding: 0 }}>
                 <IconOrImage icon={item.icon} image={item.image} title={item.title} />
-                {item.value}
+                {`${item.value}`}
               </td>
             </tr>
           ))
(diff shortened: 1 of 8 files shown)


src/textual_summary/generic_table_row.jsx
```diff
@@ -22,7 +22,7 @@ const renderMultivalue = function renderMultivalue(values, onClick) {
 };
 
 const renderValue = function renderValue(value, onClick) {
```
Hyperkid123: Just a small nitpick. You don't have to write `const functionName = function functionName(...)`, you can replace it with `const functionName = (...) => {}`. The function definition after the `=` is useless. I have noticed that you are using this very often.

Hyperkid123:
Seems to be working just fine. Bool values are now visible :)

src/textual_summary/generic_table_row.jsx
karelhala: Or you can write just `function renderValue(value, onClick) {` which is probably the best, if it's not just simple function with one liner as return otherwise go ahead and use `const functionName = (...) => ({some: object})`

martinpovolny:
@karelhala, @himdel : can you, please, merge this one?

@karelhala : will you, please, also release a new version with this?

PR #32, opened 2018-05-28
Mock functions for i18n

Hyperkid123:
Created a simple `translate` and `sprintf` function, that should act as a proxy for translation. These functions will work only in environment, where `__` and `sprintf` functions are available on window (as in ManageIQ). If they are not available, it will return original string or mask.

### DIFF
--- a/demo/demo-app.jsx
+++ b/demo/demo-app.jsx
@@ -3,6 +3,7 @@ import * as React from 'react';
 import { Form, Field } from 'react-final-form';
 import { Form as PfForm, Col, Row, Button, Grid } from 'patternfly-react';
 import { FinalFormSwitch } from '../src/forms';
+import { translate, sprintf } from '../src/global-functions';
 
 const onSubmit = values => console.log('onSubmit: ', values);
 
@@ -30,7 +31,7 @@ const wrapperComponent = () => (
           </Row>
           <Row>
             <Col xs={12}>
-              <Button type="submit">Submit</Button>
+              <Button type="submit">{translate(sprintf('Mask $d'), 21)}</Button>
             </Col>
           </Row>
         </PfForm>
--- a/demo/demo-app.jsx
+++ b/demo/demo-app.jsx
@@ -3,7 +3,6 @@ import * as React from 'react';
 import { Form, Field } from 'react-final-form';
 import { Form as PfForm, Col, Row, Button, Grid } from 'patternfly-react';
 import { FinalFormSwitch } from '../src/forms';
-import { translate, sprintf } from '../src/global-functions';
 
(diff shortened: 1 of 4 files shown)


Hyperkid123:
@martinpovolny can you check it out?

martinpovolny:
@mzazrivec : what do you think?

himdel:
This doesn't work.

Gettext needs that `__(...)` to find the strings to extract.

Wrapping that in a helper function just means we will have no strings..

himdel:
(`sprintf` is just fine, but the "mock" `translate` needs to be called `__`, not `translate`)

mzazrivec:
@Hyperkid123 The `translate()` routine needs to be renamed to `__()`. We don't want to be using something special here.

Hyperkid123:
@mzazrivec ok i will rename it
@himdel the helper function seems to be working just fine. I have just tested it and this is the result. This source 
```
import React from 'react';
import { translate, sprintf } from '../global-functions';

export const TranslateDemo = () => (
  <div>
    <h1>Translated string 1: {sprintf(translate('%d%% in use of %d total'), 15, 50)}</h1>
    <h1>Translated string 2: {translate('Submit')}</h1>
    ## next line won't work in react-ui-components dev server
    <h1>Translated string 3: {__('Submit')}</h1>
  </div>
);
```
will render this is manageIQ:

screenshot from 2018-06-07 09-33-32
Is there something i am missing?

demo/demo-app.jsx
```diff
@@ -3,6 +3,7 @@ import * as React from 'react';
 import { Form, Field } from 'react-final-form';
 import { Form as PfForm, Col, Row, Button, Grid } from 'patternfly-react';
 import { FinalFormSwitch } from '../src/forms';
+import { translate, sprintf } from '../src/global-functions';
```
mzazrivec: This should read `import { __, sprintf } ...`, right?

Hyperkid123: @mzazrivec oh this file should not even be in this PR. Thanks

martinpovolny:
> Is there something i am missing?

Your stuff works in run time, but you are missing the part where the strings are collected for translation. This is done by parsing the sources and collecting all the calls to `__(args)` and a several other functions.

Hyperkid123:
@martinpovolny yes @himdel already explained.

Hyperkid123:
@martinpovolny i have renamed the translate function to `__(...)` and tested it in ManageIQ (even non existing keys) and it seems to be working. When this is merged i think i have all i need to start working on the provider forms.

Hyperkid123:
@martinpovolny @karelhala @himdel can we get this merged? Or is there something else you want me to do?

PR #33, opened 2018-05-31
Snapshot form col fix

Hyperkid123:
Replaced single checkbox witch switch component and fixed its horizontal alignment.

I have also encountered a issue with react-test-renderer. It was creating a error `Uncaught TypeError: Cannot read property 'offsetWidth' of null`, while i was trying to create a snapshot of the Switch component. Have no idea what was causing it. We don't actually need that renderer, we can use `shallow` and `mount` from enzyme and some serializer. I have added `enzyme-to-json` package and that works just fine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "css-loader": "^0.28.9",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
+    "enzyme-to-json": "^3.3.4",
     "eslint": "^4.17.0",
     "eslint-config-airbnb": "^16.1.0",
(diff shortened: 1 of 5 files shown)


Hyperkid123:
@miq-bot assign @karelhala

karelhala:
Tests are OK, storybook is good! @Hyperkid123 will you take the follow up PR where you'd rewrite tests to use `enzyme-to-json`?

Hyperkid123:
@karelhala yes i will. Can you please make a new release when this gets merged? @hstastna tested the component in UI and it works, so i would like to get it in.

PR #34, opened 2018-06-01
Tagging changes

PanSpagetka:
Delete local style to avoid redundancy, same styles are(will be) defined in patternfly when https://github.com/patternfly/patternfly-react/pull/351 is merged.

On `LOAD_STATE` action add `affectedItems` into state, to be consistent with `RESET_STATE` action


src/tagging/components/taggingWithButtons.jsx
```diff
@@ -10,6 +10,7 @@ class TaggingWithButtons extends React.Component {
   onTagDeleteClick = (tagCategory, tagValue) => this.props.onTagDeleteClick({ tagCategory, tagValue });
 
   render() {
+    console.log('render with buttons');
```
karelhala: Please remove `console.log`.

karelhala:
There seems to be couple of problems when running the storybook, it looks like you forgot to initialize state with `affectedItems`. Simplest solution would be to add `affectedItems: (state = {}) => state,` into src/tagging/reducers/index.js on line 8 to combineReducers.

Also some default values for storybook looks like they are off (expecting `object` but recieved `string`) can you take a look at it?

PR #35, opened 2018-06-04
TextualSummary: display nulls as empty strings

martinpovolny:
Partial fix for: https://github.com/ManageIQ/manageiq-ui-classic/issues/4013

Plus more tests and storybooks.

### DIFF
--- a/src/textual_summary/generic_table_row.jsx
+++ b/src/textual_summary/generic_table_row.jsx
@@ -2,6 +2,8 @@ import * as React from 'react';
 import PropTypes from 'prop-types';
 import IconOrImage from './icon_or_image';
 
+const filterValue = val => val == null ? '' : String(val);
+
 const renderMultivalue = function renderMultivalue(values, onClick) {
   return (
     <table cellPadding="0" cellSpacing="0">
@@ -11,7 +13,7 @@ const renderMultivalue = function renderMultivalue(values, onClick) {
             <tr onClick={e => onClick(item, e)} key={i} className={item.link ? '' : 'no-hover'} title={item.title}>
               <td style={{ border: 0, margin: 0, padding: 0 }}>
                 <IconOrImage icon={item.icon} image={item.image} title={item.title} />
-                {`${item.value}`}
+                {filterValue(item.value)}
               </td>
             </tr>
           ))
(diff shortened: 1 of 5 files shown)


src/textual_summary/generic_table_row.jsx
```diff
@@ -2,6 +2,8 @@ import * as React from 'react';
 import PropTypes from 'prop-types';
 import IconOrImage from './icon_or_image';
 
+const filterValue = val => val == null ? '' : val;
```
himdel: ...`? '' : String(val)` and we could use `{filterValue(item.value)}` instead of `{\`${filterValue(item.value)}\`}`

martinpovolny: Will do. Thx!

himdel:
Thanks :)

Otherwise, LGTM, `== null` catches both `null` and `undefined`, sounds about right :)

Will test and merge in the morning if nobody gets there faster :)

src/textual_summary/generic_table_row.jsx
karelhala: This can be further simplified into
```javascript
const filterValue = val => val || '';
```
No need to check for null, because `null == undefined` only if you'd wan to strictly check val to null or undefined.
```javascript
filterValue(false)
"" //output
filterValue([])
[]  //output
filterValue({})
{}  //output
filterValue(undefined)
""  //output
filterValue(null)
""  //output
```

src/textual_summary/tests/__snapshots__/textual_summary.test.js.snap
```diff
@@ -0,0 +1,496 @@
...
+            <span>
+               
+              Amazon EC2 Network
+            </span>
+          </td>
+        </tr>
+        <tr
+          className="no-hover"
+          onClick={[Function]}
+          title={undefined}
```
karelhala: This is from previous version of your commit right?

karelhala:
Tests are failing, @martinpovolny please update your snapshots.

src/textual_summary/generic_table_row.jsx
himdel:
```js
`filterValue(0)`
"" // output
```

not really ;)

Hyperkid123:
@karelhala we need travis :D

src/textual_summary/generic_table_row.jsx
karelhala: Yeah that's correct, so if we want to print `false` and `0` in these places I am fine with it as it is now.

martinpovolny:
@karelhala : fixed the tests. Thanks, guys!

martinpovolny:
ping @himdel ?

himdel:
Sorry, merged :) No milestones here?

martinpovolny:
Right, we should start cutting milestones the same way we do elsewhere. I think that @Fryguy has a script that adds milestone to a list of repositories. @Fryguy: can you, please, include this repo?

@karelhala : can yoiu do a release for me, please?

PR #36, opened 2018-06-05
Update packages and use webpack 4

karelhala:
All packages updated, webpack config changed to follow v4. No UI changes or anything, just build.

https://5b16ab7db3127433dd13bbab--sleepy-spence-62a12c.netlify.com


priley86:
i like the approach w/ netlify b.t.w. ;) - i think we will be using the same path in PF React soon...

Hyperkid123:
@karelhala just one question. Why do we want alpha packages? 

Also, storybooks os netlify are broken. Something is requesting jQuery `Uncaught Error: Cannot find module 'jquery'`

priley86:
this will unblock our QE team on testing latest v2v UI. It's much appreciated if we can get this in soon...
https://github.com/ManageIQ/miq_v2v_ui_plugin/issues/363

cc: @bthurber @Yadnyawalkya

AparnaKarve:
@karelhala How do you suggest we test this locally?

Will this work in classic-ui package.json  - 
```
"@manageiq/react-ui-components": "https://github.com/karelhala/react-ui-components#update/pf-react-v2",
```

package.json
```diff
@@ -49,67 +49,69 @@
...
+    "style-loader": "^0.21.0",
+    "url-loader": "^1.0.1",
+    "webpack": "^4.10.2",
+    "webpack-cli": "^3.0.2",
+    "webpack-dev-server": "^3.1.4"
   },
   "dependencies": {
-    "final-form": "^4.6.1",
-    "patternfly-react": "^1.16.4",
+    "extract-text-webpack-plugin": "^4.0.0-beta.0",
```
himdel: Was this one intentional?

karelhala: I tried it, but should be removed, good catch!

karelhala:
@Hyperkid123:
> just one question. Why do we want alpha packages?

Webpack is improving way too fast compared to storybook, but they released alpha version.

@AparnaKarve:
> Will this work in classic-ui package.json -

I don't think that it's going to work. Since we publish build into npm and not into github.

karelhala:
@Hyperkid123 great catch! Removing jquery from provide plugins works and netlify does not complain about it anymore.

AparnaKarve:
> I don't think that it's going to work. Since we publish build into npm and not into github. If you want to test it you'd have to download this package and build it locally.

@karelhala Thanks. I will just wait until it's merged... hopefully today :)

scripts/storybook/webpack.config.js
```diff
@@ -15,18 +15,34 @@ module.exports = {
...
+              },
             },
-          ],
-        }),
+          },
+        ],
+        enforce: 'pre',
+      },
+      {
+        test: /\.s?[ac]ss$/,
```
himdel: Not sure if `/\.(sa|sc|c)ss$/` wouldn't be safer - not sure any of those loaders have `.ass` file support.

karelhala: will change it

karelhala:
@himdel Provide plugin kinda broke our builds on netlify because it exposed jquery and other variables, which is unwanted. And we didn't really used it anyway.

PR #38, opened 2018-06-07
Add custom commit verify and trigger minor release <?.x.?>

karelhala:
### Add support for semantic release
* When PR is merged all commits which were not released are checked for release string
* Once new version is calculated build is triggered and release is triggered

### Release strings
Not that these strings can be in any part of commit message.
* Major - `<x.?.?>`, `<x.x.?>`, `<x.x.x>`, `<x.?.x>`, 
* Minor - `<?.x.?>`, `<?.x.x>`,
* Patch - `<?.?.x>`,
* No release - `<no>`, `<No>`, `<nO>`, `<NO>`

### Prerequisits
We need to enable travis to run these tasks with these variables `GH_TOKEN` or `GITHUB_TOKEN` and `NPM_TOKEN`

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,9 @@
     "doc": "react-docgen",
     "storybook": "start-storybook -c scripts/storybook -p 6006",
     "build-storybook": "build-storybook -c scripts/storybook -o .out",
-    "coveralls": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js"
+    "coveralls": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
+    "prepublishOnly": "npm run build",
+    "release": "semantic-release"
   },
   "engines": {
     "node": ">= 6.0.0",
@@ -56,6 +58,7 @@
     "@storybook/addon-storysource": "^3.4.6",
     "@storybook/addons": "^4.0.0-alpha.0",
     "@storybook/react": "^4.0.0-alpha.0",
+    "@khala/commit-analyzer-wildcard": "^1.0.4",
     "autobind-decorator": "^2.1.0",
     "babel-core": "^6.26.3",
     "babel-eslint": "^8.2.1",
@@ -97,7 +100,8 @@
     "url-loader": "^1.0.1",
     "webpack": "^4.10.2",
     "webpack-cli": "^3.0.2",
-    "webpack-dev-server": "^3.1.4"
+    "webpack-dev-server": "^3.1.4",
+    "semantic-release": "^15.5.0"
   },
   "dependencies": {
     "final-form": "^4.7.3",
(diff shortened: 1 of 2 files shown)


scripts/verify-commits.js
```diff
@@ -0,0 +1,42 @@
+const releaseTypes = ['major', 'minor', 'patch', 'prerelease'];
+let patterns = {
+  major: /<x\.[x|?]\.[x|?]>/,
+  minor: /<\?\.x\.[x|?]>/,
+  patch: /<\?\.\?\.x>/,
+  noRelease: /<[N|n][O|o]>/,
```
himdel: this could be written as `/<no>/i` :)

package.json
```diff
@@ -112,5 +115,12 @@
     "redux": "^3.7.2",
     "redux-form-validators": "^2.4.3",
     "redux-mock-store": "^1.5.1"
+  },
+  "release": {
+    "analyzeCommits": [
```
himdel: Is this specific to our own commit analyzer?

Because, looking at https://github.com/semantic-release/commit-analyzer#options , it looks like analyzeCommits is always an object, not an array..

scripts/verify-commits.js
```diff
@@ -0,0 +1,42 @@
...
+};
+
+async function commitAnalyzer(pluginConfig, { options: { analyzeCommits }, commits, logger }) {
+  let releaseNumber = 3;
+  analyzeCommits.map((onePlugin) => {
+    patterns = {
```
himdel: if this gets run multiple times, patterns will keep growning...

But perhaps more to the point, what does this do? Is the idea here that we can override the patterns in the config?

himdel:
I think LGTM :) Should we wait for the variables or will this just keep not doing anything when those are missing?

scripts/verify-commits.js
karelhala: Great catch! I changed it so patterns can come from package.json as well.

package.json
karelhala: Their documentation sucks, it can either be object or array of objects.

scripts/verify-commits.js
karelhala: This is to enable options to override patterns from package.json or .releaserc file, we could however we need to provide some default patterns. You are correct that this might keep growing, but this script is runned only once. And then release is triggered on non `master` branch this script isn't runned at all.

karelhala:
> will this just keep not doing anything when those are missing?

It will just fails so no need to wait for those variables.

scripts/verify-commits.js
```diff
@@ -0,0 +1,42 @@
...
+  });
+  logger.log(`
+  Using patterns:
+      major - ${patterns.major}
+      minor - ${patterns.minor}
+      patch - ${patterns.patch}
+      noRelease - ${patterns.noRelease}`);
+  let i = 0; const iMax = commits.length;
+  for (; i < iMax; i++) {
+    if (commits[i].message.search(new RegExp(patterns.major), i) !== -1) {
```
himdel: , 'i')

`i` is a number :)

PR #42, opened 2018-06-16
Textual improvements

martinpovolny:
* Support links in SimpleTable
 * SimpleTable: Do not break words in table heading.

### DIFF
--- a/src/textual_summary/simple_table.jsx
+++ b/src/textual_summary/simple_table.jsx
@@ -17,32 +17,37 @@ export default function SimpleTable(props) {
     return label;
   };
 
-  const renderValue = (value, j) => {
+  const renderValue = (value, j, onClick) => {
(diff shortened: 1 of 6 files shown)


src/textual_summary/simple_table.jsx
```diff
@@ -17,32 +17,37 @@ export default function SimpleTable(props) {
...
+      // expandable value
       return (
         <td key={j}>
           <div className="pull-left expand">{value.value}</div>
           <div className="pull-right" />
         </td>
       );
+    } else if ((value != null) && (typeof value === 'object') && value.link) {
+      // value with a link
+      return <td onClick={e => onClick(value, e)} key={j}>{value.value}</td>;
```
Hyperkid123: Maybe we could add `cursor: pointer` style to clickable cells?

PR #44, opened 2018-06-25
Added i18n to snapshot form component.

Hyperkid123:
Removed props for form labels and error messages and replaced them with gettext functions.

### DIFF
--- a/src/vm-snapshot-form/snapshot-form/snapshotForm.jsx
+++ b/src/vm-snapshot-form/snapshot-form/snapshotForm.jsx
@@ -5,15 +5,14 @@ import PropTypes from 'prop-types';
 import { required } from 'redux-form-validators';
 
 import { renderFinalFormField, renderFinalFormTextArea, FinalFormSwitch } from '../../forms/';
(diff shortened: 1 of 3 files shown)


src/vm-snapshot-form/snapshot-form/snapshotForm.jsx
```diff
@@ -26,36 +25,36 @@ export const VmSnapshotForm = ({
           <Col xs={12}>
             {!hideName &&
               <Field
-                validate={nameRequired && required({ msg: errorMessages.name })}
+                validate={nameRequired && required({ msg: __('Required') })}
```
ohadlevy: does it makes sense to ensure the strings are similar to the rails ones? so we:
1. don't create new strings
2. consistent with the API and rails based UI ?

Hyperkid123: Yes. It is created to match the manage-ui-classic. You can check #32 for mode info. Strings are taken directly from the previous form.

PR #46, opened 2018-06-25
Fixed single-quotes error.

Hyperkid123:
Removed small bug that prevented build.


Hyperkid123:
cc @martinpovolny can ve get this in quickly?

Hyperkid123:
@Fryguy can we enable travis for this repo? We could avoid bugs like this getting in.

Issue #47, opened 2018-06-25
Storybook is broken


PanSpagetka:
@karelhala when try to run `yarn storybook` it blows up, there are some issues with path resolution. Probably caused by #36 

when reverting storybook and webpack changes, storybook works again

PR #48, opened 2018-06-25
Fix storybook build by extending default build config

karelhala:
### Fixes https://github.com/ManageIQ/react-ui-components/issues/47
Storybook task is broken because of wrong webpack config. Looks like storybook for webpack 4 does not imports rules from master config so we need to reimport them by ourselves.

When doing so let's ignore node_modules, data, stories folders and vendor file for eslint.


Hyperkid123:
Works just fine. Ty.

PR #49, opened 2018-06-26
Rbac users table

Hyperkid123:
Created a simple sorting table as a par of rbac user editing components.

Strybook is available on https://5b31fa4967610c6fac9c340d--confident-banach-ee0a7a.netlify.com/?selectedKind=Rbac%20forms&selectedStory=Rbac%20users%20table&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel


src/rbac-forms/rbacUsersTable.jsx
```diff
@@ -0,0 +1,150 @@
+import React, { Component } from 'react';
+import { Table, Checkbox, Icon } from 'patternfly-react';
+import PropTypes from 'prop-types';
+import './styles.scss';
+
+class RbacUsersTable extends Component {
```
karelhala: This component looks great! We don't have to limit ourselves only to rbac and use it in every table since this table has same signature as any table in MiQ. Allow enabling of icons and selectors plus let consumer pass icons/images in first column.

I'd consider moving this to seperate pack perhaps `src/table` so it can be used seperately from forms.

Hyperkid123:
@karelhala component was renamed and moved.

PR #53, opened 2018-06-27
[WIP ]Rbac user editing, and routing.

Hyperkid123:
I have made some modifications to enable users editing via RbacUserForm.

I have also created a demo which includes all rbac components, that are needed for adding/viewing/editing users while using client-side redux routing. It should be reproducible in manageiq-ui-classic and with some API tuning we could replace bunch of old code.

Demo is using hashHistory, so it should not collide with any ruby routing.

Storybook for demo: https://5b34e0ddfdd72a0b134bf4fb--confident-banach-ee0a7a.netlify.com/?selectedKind=Rbac%20forms&selectedStory=Rbac%20complex%20demo&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel

### DIFF
--- a/src/manageiq-validators/manageIqPropTypes.js
+++ b/src/manageiq-validators/manageIqPropTypes.js
@@ -1,0 +1,15 @@
+import { emailPattern } from './';
+
+export const emailPropType = ({
+  props,
+  propName,
+  componentName,
+  isRequired,
+}) => {
+  if (!props[propName] && !isRequired) {
+    return undefined;
+  }
+  return emailPattern.test(props[propName])
+    ? undefined
+    : new Error(`Invalid prop  ${propName} supplied to ${componentName} Validation failed. Expected email address.`);
+};
--- a/src/rbac-forms/rbacUserManagementModule/index.jsx
+++ b/src/rbac-forms/rbacUserManagementModule/index.jsx
@@ -2,7 +2,7 @@ import React, { PureComponent } from 'react';
 import { Provider, connect } from 'react-redux';
 import { Route, Switch } from 'react-router-dom';
 import { ConnectedRouter, goBack } from 'connected-react-router';
-import { Button, Grid, Row, Col, ListGroup, ListGroupItem } from 'patternfly-react';
+import { Button, Grid, Row, Col, ListGroup, ListGroupItem, Spinner } from 'patternfly-react';
 import PropTypes from 'prop-types';
 import { history, store } from './redux/reducers/';
 import { navigate, requestUsers, deleteUser } from './redux/actions/actions';
@@ -86,8 +86,8 @@ class App extends PureComponent {
   renderuserEdit = props => <UserEdit onSave={this.mockUserSave} {...props} />;
 
   render() {
-    if (!this.props.isLoaded) {
-      return <div>Loading</div>;
+    if (this.props.isFetching || !this.props.isLoaded) {
+      return <div><Spinner loading size="lg" /></div>;
     }
     return (
       <Grid fluid>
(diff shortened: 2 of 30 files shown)


Hyperkid123:
@karelhala sorry for adding demo here, i've noticed it after i've pushed it...

Hyperkid123:
@karelhala @himdel @martinpovolny Please check the storybook in description (and code). I would like to know your opinion on this. Good/bad idea? All components can be user separately, so if this turns out as a horrible idea, we can still use these components for the forms etc.

martinpovolny:
@PanSpagetka, @karelhala : can you, please, test this one?

PanSpagetka:
I have issues when i try to run it in Firefox (60.0.2 ), when I try chrome it is ok

Screenshot:
firefox_screenshot_2018-07-11t10-16-40 135z

src/manageiq-validators/manageIqPropTypes.js
```diff
@@ -0,0 +1,15 @@
...
+  propName,
+  componentName,
+  isRequired,
+}) => {
+  if (!props[propName] && !isRequired) {
+    return undefined;
+  }
+  return emailPattern.test(props[propName])
+    ? undefined
+    : new Error(`Invalid prop  ${propName} supplied to ${componentName} Validation failed. Expected email address.`);
```
PanSpagetka: Maybe use the `__('bla')__`

src/rbac-forms/rbacUserManagementModule/index.jsx
```diff
@@ -0,0 +1,137 @@
...
+    <Button disabled={!canEdit} onClick={() => handleEdit(editUserId)}>Edit selected</Button>
+    <Button disabled={!canEdit} onClick={() => handleDelete(editUserId)}>Delete selected</Button>
+  </div>
+);
+
+Toolbar.propTypes = {
+  handleAdd: PropTypes.func.isRequired,
+  handleEdit: PropTypes.func.isRequired,
+  canEdit: PropTypes.bool.isRequired,
+  editUserId: PropTypes.number,
```
PanSpagetka: Shouldn't this be also required?

Hyperkid123: Probably yes, but the index.jsx is just a demo file for storybook. This wrapping component should be implemented in MIQ and just provide data and callbacks for other components.

Hyperkid123:
I have encountered an issue with react-router-dom package in ui-classic. It seems when you add `<Router />` component and render the component it will start an infinite loop... We (me and @karelhala) will look into it and see what can be done about it. If we cannot fix it, the whole idea of client-side routing in ui-classic will be useless and this can be scrapped.

src/manageiq-validators/manageIqPropTypes.js
Hyperkid123: Oh you mean for the error messages. Nah it will be visible only in console, this is a message for developers not users. Its just a custom prop error message.

Hyperkid123:
The issue with routing in miq was resolved. See https://github.com/ManageIQ/manageiq-ui-classic/pull/4284. Thanks @karelhala and @himdel for help. It turned out to be very easy fix.

Hyperkid123:
@PanSpagetka the firefix issue is caused by redux dev tools... Removing `window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__()` fixes the issue. I will keep it there for now. It will not make it the final component anyway.

PR #54, opened 2018-07-02
Fix overflow of security group form in modals

Hyperkid123:
Added fluid property to form wrapper.


Hyperkid123:
@martinpovolny fix for https://github.com/ManageIQ/manageiq-ui-classic/pull/4231

PR #58, opened 2018-07-12
Disabled property for FinalForm inputs

Hyperkid123:
Added disabled prop and missing styling for FinalFormFields.

### DIFF
--- a/src/forms/style.scss
+++ b/src/forms/style.scss
@@ -60,6 +60,27 @@ $input-default-box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.075);
   }
 }
 
+.final-form-select.is-disabled {
+  .Select-control {
+    cursor: not-allowed;
+    background-image: none;
+    box-shadow: none;
+    .Select-placeholder {
+      color: $color-pf-black-500;
+    }
+    &:hover {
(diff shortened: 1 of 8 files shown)


src/forms/style.scss
```diff
@@ -60,6 +60,27 @@ $input-default-box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.075);
   }
 }
 
+.final-form-select.is-disabled {
+  .Select-control {
```
ohadlevy: just wondering, why do we need special css? isnt it by default by bootstrap/patternfly when the disable attribute is set?

Hyperkid123: apparently no. Adding just the disabled prop has no effect on the component.

ohadlevy:
I wonder if we are missing anything? as when we did the same for
redux-forms there was no need to add additional css? (nor do i think i had
to do it before)...

Hyperkid123:
@ohadlevy what component did u use? This implementation uses ReactSelect. When i was creating these components there was no select component in patternfly-react.

Hyperkid123:
@karelhala

Hyperkid123:
@miq-bot assign @karelhala

PR #59, opened 2018-07-12
FinalFormSelect onChange event now returns only value from option.

Hyperkid123:
Just a small change of selects. Not it will return only the actual value of selected option. This should make building forms a bit easier, because you only need values most of the time.

### Single select
`stateValue: { value: 55, label: 'Option label' }` => `stateValue: 55`
### Multi select
`stateValue: [{ value: 1, label: 'First option' }, { value: 2, label: 'Second option' }]` => `stateValue: [1, 2]`

Should not break any forms just yet.


Hyperkid123:
@karelhala

Hyperkid123:
It might be useful sometimes but i think that in 90% use case you would rather want an object of values that you can directly send to endpoint. With the current way you would have to parse the form values before sending them. And you have the full options list available anyway.

karelhala:
Okay, sounds good enough, merging.

PR #60, opened 2018-07-17
Tag Category line wrapping

PanSpagetka:
Make Tag Category to wrap long lines.

### Screenshot:
screenshot from 2018-07-17 11-44-10

@karelhala

### DIFF
--- a/src/tagging/sass/_tag-category.scss
+++ b/src/tagging/sass/_tag-category.scss
@@ -9,8 +9,10 @@
   margin-bottom: 5px;
   font-size: 14px;
   line-height: 24px;
+  display: -ms-flexbox;
   display: flex;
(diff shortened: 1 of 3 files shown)


src/tagging/sass/_tag-category.scss
```diff
@@ -9,7 +9,8 @@
   margin-bottom: 5px;
   font-size: 14px;
   line-height: 24px;
-  display: table;
+  display: flex;
```
karelhala: Please add `ms` specific flex properties so we can use it in IE11.
```css
  display: -ms-flexbox;
  display: flex;
  -ms-flex-wrap: wrap;
      flex-wrap: wrap;
```

PR #65, opened 2018-07-24
Fixed amazon-provider-form data format on spy update.

Hyperkid123:
Fixed the data format to match https://github.com/ManageIQ/manageiq-providers-amazon/pull/468


Hyperkid123:
@miq-bot assign @martinpovolny

PR #66, opened 2018-07-30
Max length form field property

Hyperkid123:
Added property to enable maxLength on final form fields.


Hyperkid123:
@miq-bot assign @karelhala

PR #67, opened 2018-08-06
Use something other than `about:blank` as testURL

miha-plesko:
With this commit we prevent command

```
yarn test
```

from failing for all the tests with recent enough jest due to https://github.com/jsdom/jsdom/issues/2304. Problem is that Redux has some issues with `about:blank` address so everything explodes.

@miq-bot add_label enhancement
@miq-bot assign @Hyperkid123

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,8 @@
     ],
     "moduleNameMapper": {
       "\\.(css|scss)$": "identity-obj-proxy"
-    }
+    },
+    "testURL": "http://localhost"
   },
   "author": "",
   "license": "Apache-2.0",


Hyperkid123:
@miha-plesko are there some steps to reproduce? Or do you have some error log? We did not had this issue before.

miha-plesko:
I can reproduce it by checking out master and running `yarn test`:

```bash
$ yarn test
yarn run v1.4.0
$ jest
 FAIL  src/textual_summary/tests/tag_group.test.js
  ● Test suite failed to run

    SecurityError: localStorage is not available for opaque origins

      at Window.get localStorage [as localStorage] (node_modules/jsdom/lib/jsdom/browser/Window.js:257:15)
          at Array.forEach (<anonymous>)

 FAIL  src/rbac-forms/tests/rbacUserPreview.test.jsx
  ● Test suite failed to run

    SecurityError: localStorage is not available for opaque origins

      at Window.get localStorage [as localStorage] (node_modules/jsdom/lib/jsdom/browser/Window.js:257:15)
          at Array.forEach (<anonymous>)

 FAIL  src/textual_summary/tests/multilink_table.test.js
  ● Test suite failed to run

    SecurityError: localStorage is not available for opaque origins

      at Window.get localStorage [as localStorage] (node_modules/jsdom/lib/jsdom/browser/Window.js:257:15)
          at Array.forEach (<anonymous>)

...
```

Where my yarn is

```
$ yarn -v
1.4.0
```

and my jest is `jest@23.4.2`

Hyperkid123:
did you run `yarn vendor` command?

miha-plesko:
I did now. I also did `yarn remove jest` and then

```
$ yarn vendor
$ yarn install
```

again but the error persists.. Am I doing something wrong, because after `yarn vendor` it even fails with "jest not found" error so I have to do `yarn install` as well?

Hyperkid123:
well if you did remove the jest its obvious it will fail.
basically to setup the repo you have to run these commands: `yarn; yarn vendor;` and you should be set up.

What version of node do you have?
```
$ node -v
v8.11.3
$ yarn -v
1.7.0
```

Hyperkid123:
That said i am not against this addition, but i would like to explore all possible causes before we will add this, because we did not had this issue so far.

Hyperkid123:
@karelhala ok so it seems that on fresh copy of the repo you will get this error. This fixes the issue and i would recommend to get this in.

Hyperkid123:
@miq-bot assign @karelhala

Hyperkid123:
@miq-bot add_label bug

karelhala:
No need to release new version, since it's just test config.

PR #68, opened 2018-08-13
Added option to render children to final form components.

Hyperkid123:
Sometimes it might be required to add additional components to form components (e.q. change password link in miq)


Hyperkid123:
@miq-bot assign @karelhala

PR #69, opened 2018-08-13
User form editing

Hyperkid123:
Modification to RbacUserForm component which enable editing some user.

### Adding user
add `newRecord` prop to component to signal adding

### Editing
add `initialValues` prop to edit some user

depends on #68

### DIFF
--- a/src/rbac-forms/formPasswordFragment.jsx
+++ b/src/rbac-forms/formPasswordFragment.jsx
@@ -1,0 +1,95 @@
+import React, { PureComponent, Fragment } from 'react';
+import { Field } from 'react-final-form';
+import { required } from 'redux-form-validators';
+import { Col } from 'patternfly-react';
+import PropTypes from 'prop-types';
+import { FinalFormField } from '../forms';
+import { __ } from '../global-functions';
+
+export default class PasswordFragment extends PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = {
+      editEnabled: false,
+    };
+  }
+
+  handleChangeEdit = () => this.setState(prevState => ({ editEnabled: !prevState.editEnabled }));
+
+  render() {
+    const { isEditing, changeValue } = this.props;
+    const { editEnabled } = this.state;
+    if (!isEditing || editEnabled) {
+      return (
+        <Fragment>
+          <Col xs={12}>
+            <Field
+              id="password"
+              name="password"
+              type="password"
+              validate={required({ msg: __('Required') })}
+              component={FinalFormField}
+              label={__('Password')}
+            >
+              {
+                isEditing &&
+                <Col md={2}>
+                  <button
+                    id="password-change-disabler"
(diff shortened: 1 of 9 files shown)


Hyperkid123:
cc @PanSpagetka

Hyperkid123:
@miq-bot remove-label unmergeable

Hyperkid123:
@miq-bot assign @karelhala

src/rbac-forms/formPasswordFragment.jsx
```diff
@@ -0,0 +1,95 @@
...
+    const { editEnabled } = this.state;
+    if (!isEditing || editEnabled) {
+      return (
+        <Fragment>
+          <Col xs={12}>
+            <Field
+              id="password"
+              name="password"
+              type="password"
+              validate={required({ msg: __('Required') })}
```
karelhala: You should be able to pass custom validators for password as well.

src/rbac-forms/formPasswordFragment.jsx
```diff
@@ -0,0 +1,95 @@
...
+              component={FinalFormField}
+              label={__('Password')}
+            >
+              {
+                isEditing &&
+                <Col md={2}>
+                  <button
+                    id="password-change-disabler"
+                    tabIndex="-1"
```
karelhala: Why the negative `tabIndex`? You will not be able to access this button when clicking on `tab` button.

src/rbac-forms/formPasswordFragment.jsx
```diff
@@ -0,0 +1,95 @@
...
+        <Field
+          name="password"
+          type="password"
+          component={FinalFormField}
+          label={__('Password')}
+          disabled
+          placeholder="●●●●●●●●"
+        >
+          <Col md={2}>
+            <button id="password-change-enabler" tabIndex="-1" type="button" className="button-link" onClick={this.handleChangeEdit}>
```
karelhala: Again why the negative `tabIndex`?

src/rbac-forms/formPasswordFragment.jsx
Hyperkid123: Honestly can remember, will remove it

src/rbac-forms/formPasswordFragment.jsx
Hyperkid123: OK this one is because of the disabled property. When tabbing through the fields, i found it confusing when the disabled password field got targeted. I was expecting in to "jump" to the next field in form.

karelhala: Oh I see, but once the password is changed the button should be enabled and possible for targeting, am I right?

Hyperkid123: Yes and it is because it will render different inputs and those does not have the tabindex prop.

Hyperkid123:
@karelhala btw this on depends on #68

PR #70, opened 2018-08-13
Modified RbacUserPreview props to match miq types.

Hyperkid123:
Change prop type form user id from number to string.


Hyperkid123:
cc @PanSpagetka

PR #71, opened 2018-08-13
Update generic preview table

Hyperkid123:
Added component update when it receives new data. For now its only checking for data length, eventualy it should compare event the content. (NOTE this is here because react tabular. I would strongly recommend create our own implementation in near future for GTL components)

Also added blue background for selected rows.

### DIFF
--- a/src/table/genericPreviewTable.jsx
+++ b/src/table/genericPreviewTable.jsx
@@ -37,8 +37,16 @@ class GenericPreviewTable extends Component {
     >
       {label}
     </Table.Heading>);
-  cellFormat = value => <Table.Cell className="clickable">{value}</Table.Cell>;
-  celliconFormat = () => <Table.Cell className="cell-middle clickable"><Icon type={this.props.icon.type} name={this.props.icon.name} /></Table.Cell>;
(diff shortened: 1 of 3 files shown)


Hyperkid123:
cc @karelhala @PanSpagetka

src/table/genericPreviewTable.jsx
```diff
@@ -97,17 +119,16 @@ class GenericPreviewTable extends Component {
     this.state.sortableColumnPropery === columnProps.column.property,
   );
 
-  handleSelected = ({ id }) => this.setState((prevState) => {
+  handleSelected = row => this.setState((prevState) => {
+    let currentRow;
     const rows = prevState.rows.map((item) => {
-      if (item.id !== id) {
+      if (item[this.props.rowKey] !== row[this.props.rowKey]) {
```
karelhala: This condition looks kinda off, maybe consider flipping the operand and write it like this
```javascript
const rows = prevState.rows.map((item) => {
  if (item[this.props.rowKey] === row[this.props.rowKey]) {
    item.selected = !item.selected;
    currentRow = {...item};
  }
  return item;
})
```

PR #74, opened 2018-08-20
Add optional background for IconOrImage with a colored wrapper div

skateman:
This will allow us to display powerstate icons as a fonticon + background combo. If the background is set, a wrapper div with the `backgrounded-icon` class is being created. 

screenshot from 2018-08-21 19-49-01

@miq-bot add_reviewer @epwinchell 
@miq-bot add_reviewer @martinpovolny

### DIFF
--- a/src/textual_summary/icon_or_image.jsx
+++ b/src/textual_summary/icon_or_image.jsx
@@ -1,11 +1,22 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
+import './styles.scss';
 
 /**
  * Render icon or an image with a title
  */
 export default class IconOrImage extends React.Component {
-  renderIcon = () => <i className={this.props.icon} title={this.props.title} />;
+  renderIcon = () => {
+    if (this.props.background) {
+      return (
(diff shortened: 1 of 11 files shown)


src/textual_summary/icon_or_image.jsx
```diff
@@ -1,11 +1,22 @@
...
 import PropTypes from 'prop-types';
+import './styles.scss';
 
 /**
  * Render icon or an image with a title
  */
 export default class IconOrImage extends React.Component {
-  renderIcon = () => <i className={this.props.icon} title={this.props.title} />;
+  renderIcon = () => {
+    if (this.props.background && this.props.background !== '') {
```
Hyperkid123: Maybe add `this.props.background.trim() !== ''` to the condition. To avoid strings containing only white spaces.

Hyperkid123:
@skateman styles go to `src/textual_summary/styles.scss`. After build, you will get css file which has the same name as the root directory of the component (in this case textual_summary.css).

skateman:
@karelhala or @himdel can you please merge this?

PR #77, opened 2018-08-29
Rbac components fix

Hyperkid123:
Added some minor fixes for the rbac components set. 

https://github.com/ManageIQ/manageiq-ui-classic/pull/4562


Hyperkid123:
@miq-bot assign @karelhala

PR #78, opened 2018-09-03
Allow custom sorting on generic table.

Hyperkid123:
This enables adding custom sorting function and pre-sorted rows from outside. Is useful for pagination, filtering etc...


Hyperkid123:
@miq-bot assign @karelhala

PR #88, opened 2018-10-15
Fix enable/disable buttons for rbac users form.

Hyperkid123:
When editing some form and changing some values that are array or object and then changing then to match initial values, the pristine flag will be `false` even though it should be `true`.

React-final-form API is no making deep compassion, but there is an open PR that should fix it. See: https://github.com/final-form/react-final-form/issues/151

Until its fixed, i've added very simple compassion.

### DIFF
--- a/src/forms/finalFormComponent.jsx
+++ b/src/forms/finalFormComponent.jsx
@@ -11,8 +11,7 @@ const componentTypes = ['radio', 'checkbox', 'textarea', 'select', 'textfield', 
 const switchComponents = ['radio', 'checkbox'];
 const inputTypes = ['text', 'email', 'number', 'password'];
 const selectValue = (option, labelKey, valueKey) =>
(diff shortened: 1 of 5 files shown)


Hyperkid123:
@miq-bot assign @karelhala

src/forms/finalFormComponent.jsx
```diff
@@ -11,8 +11,16 @@ const componentTypes = ['radio', 'checkbox', 'textarea', 'select', 'textfield',
 const switchComponents = ['radio', 'checkbox'];
 const inputTypes = ['text', 'email', 'number', 'password'];
 const selectValue = (option, labelKey, valueKey) =>
-  option.sort((a, b) => (a[labelKey].toLowerCase() > b[labelKey].toLowerCase()))
-    .map(item => item[valueKey]);
+  option.sort((a, b) => {
```
martinpovolny: What about using some build-in function?
```
"31aaa".localeCompare("2AAA",'en', {sensitivity: 'base'});
```

PR #90, opened 2018-11-12
Fix bulding vendor styles.

Hyperkid123:
This was causing issues with vendor. It was overwriting the `vendor.css` and `vendor.js` files. This caused loss of pf styles in storybooks and the demo was not working, because no externals were available from `vendor.js`

All custom styles are available automatically in `dist/styles.css` and it seems that all components are build into `styles.js` (That does not make any sense to me).


Hyperkid123:
@miq-bot assing @karelhala

Hyperkid123:
@miq-bot assign @karelhala

PR #94, opened 2019-02-20
Update tagging snapshot, again

himdel:
This is broken, the removed methods have never existed in our codebase,
they come from react-select.

~~So, this is a quick fix to make the tests pass,
but not a proper fix in that this should never even be in the snapshot.~~

~~Cc @Hyperkid123 this sounds like some kind of bad config of snapshots?~~

So, this changes the test to use `shallow` render instead of `mount`, and updates the snapshot.

This should probably happen in all the tests.


Hyperkid123:
So we i would suggest using shallow render for snapshots on the tagging component. Manually modifying snapshots won't work and will break eventually again.

Hyperkid123:
This looks legit

PR #95, opened 2019-02-20
[WIP] Use shallow snapshots

himdel:
This is just 

    perl -i -npe 's/\bmount\b/shallow/' **/*.test.*
    perl -i -npe 's/{ shallow, shallow }/{ shallow }/' **/*.test.*
    jest -u

to use shallow snapshots wherever possible.

WIP because it causes 31 failures (looks like any kind of non-snapshot testing needs `mount`),
aand because @Hyperkid123 may know about places where this would be a bad idea even for a snapshot?


src/forms/tests/__snapshots__/fieldGroup.test.jsx.snap
```diff
@@ -1,743 +1,15 @@
...
-                />
-              </ErrorBlock>
-            </div>
-          </Col>
-        </div>
-      </Col>
-    </FieldGroup>
-  </ReactFinalForm>
-</Group>
+<ReactFinalForm
```
Hyperkid123: So this snapshot test does not really tells us anything. I would try use:
```javascript
... = shallow(<Component />).dive()
```
It should render to the `<FieldGroup />` component because that is what we actually want to test it. This is a common problem with components that require some context (Store for redux, Router for Links etc.)

more info about dive: https://airbnb.io/enzyme/docs/api/ShallowWrapper/dive.html

Now if that does not help we can just get rid of snapshot test completely, because there is no point snapshoting wrapper component that only provides context.

Hyperkid123:
Also you can use shallow and the try to simulate events. So if there is a test with combo mount/simulate you can't remove mount so i would just remove snapshot test.

PR #96, opened 2019-03-01
Add extraProps attributes to form components.

doruskova:
I've started convert Report Editor into React and I want to do it gradually and for that I need to use original solution for sending form data for now. That's the reason why I need to pass data attributes to form components. 

I created a new prop called `data` which spreads all data attributes to input. I did not want pass every prop to inputs so that's why i created isolated prop.

### DIFF
--- a/src/forms/finalFormComponent.jsx
+++ b/src/forms/finalFormComponent.jsx
@@ -13,16 +13,22 @@ const switchComponents = ['radio', 'checkbox'];
 const inputTypes = ['text', 'email', 'number', 'password'];
 const selectValue = option => option.sort((a, b) => a.label.localeCompare(b.label, 'en', { sensitivity: 'base' })).map(item => item.value);
 
-const componentSelect = (componentType, { input, meta, ...rest }) => ({
-  textfield: <FormControl type={rest.type || 'text'} {...input} placeholder={rest.placeholder} />,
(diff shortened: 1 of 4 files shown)


doruskova:
@miq-bot add-reviewer @himdel 
@miq-bot add-reviewer @ZitaNemeckova 
@miq-bot add-reviewer @Hyperkid123

himdel:
> I created a new prop called `data` which spreads all data attributes to input. I did not want pass every prop to inputs so that's why i created isolated prop.

So, this is confusing...

As the PR is now, you can use `data` to add any extra props to the final component.
*And*, if you want to provide `data-*` attributes, you need to repeat the `data-` in the key.

---

So... 2 possibilities I think:

* either this should be for data attributes only, but then we should be able to do `data="{{ foo: 5 }}` and end up with `data-foo="5"` in html
* or this is just extra props, but then let's call it extraProps or something like that, to avoid confusion with data-* attributes

---

Otherwise, makes sense :+1: :)

doruskova:
thanks for your review and the second option with renaming sounds good to me.

himdel:
@doruskova We can probably close this one, right?

Since you won't need the data attributes after all?

doruskova:
@himdel 
yes, it can be closed 👍

PR #97, opened 2019-03-05
Tagging round2

PanSpagetka:
Fix issues related to new version of used components (OvelayTrigger, Select). 

Change reduces structure to match changes in ManageIQ


PanSpagetka:
ping @karelhala

PR #107, opened 2019-04-29
Trigger build <?.?.x>

karelhala:


himdel:
Why? Didn't this happen automatically?

karelhala:
There was a `<no>` tag in history and if such string happen any subsequent pushes to master have automatic release turned off. You need to enable it by at least one push with magic string happen and automatic bug releases are turned back on again.

PR #109, opened 2019-05-03
Update styles of select

doruskova:
I changed styles of selects because I needed it for the Report Editor. 
Martin told me where I will find them. So, I copied styles from the link below.
https://github.com/data-driven-forms/react-forms/pull/14

And I had to move it to another file called pf-select.jsx because the code was too long. 

## Before
Snímek obrazovky 2019-05-05 v 15 13 05

## After
Snímek obrazovky 2019-05-05 v 15 09 26


doruskova:
@miq-bot assign @martinpovolny

martinpovolny:
@doruskova : can you, please, add before/after screenshots?

Also, please, update the snapshots (at best in a separate commit) to make Travis green. Thx!

Hyperkid123:
@martinpovolny i am looking at the snapshot errors and i think its caused by dependency version updates. It think we should update version of the `react-final-form` and `patternfly-react`. I see changes in completely unrelated components. They don't even use the select component.

PR #117, opened 2019-05-27
Replace deprecated @patternfly/patternfly-next with @patternfly/patternfly

himdel:
Same library, different name .. but updated the version to match UI classic.

(1.0.184 would also fix the BZ, but what's the point.)

---

Specifically, older version of patternfly introduce a class name clash between `pf-u*` styled by pf4, and `pf-utilization` by pf3. This was fixed to only use `pf-l-` and `pf-c-` prefixes (including a dash) in https://github.com/patternfly/patternfly-next/pull/1383.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1711913

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "webpack-dev-server": "^3.1.4"
   },
   "dependencies": {
-    "@patternfly/patternfly-next": "^1.0.42",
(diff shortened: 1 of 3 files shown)


himdel:
cc @Hyperkid123 @karelhala

PR #119, opened 2019-06-10
react-wooden-tree integration

brumik:
Added the [react-wooden-tree](https://www.npmjs.com/package/react-wooden-tree) component as re-export of the original component's redux version. Added the tiny CSS to look like the PF version and created an entry in the storybook for it. 

**Known problem - Solved**: The icons are from the font-awesome and I did not know how to include them (I did not wanted to put into the dependencies.)

@skateman


brumik:
@Hyperkid123 If I try to load font-awesome (added as dev-dep) it tells me that I need an appropriate loader for the file:

```
ERROR in ./node_modules/font-awesome/css/font-awesome.css 7:0
Module parse failed: Unexpected character '@' (7:0)
You may need an appropriate loader to handle this file type.
| /* FONT PATH
|  * -------------------------- */
> @font-face {
|   font-family: 'FontAwesome';
|   src: url('../fonts/fontawesome-webfont.eot?v=4.7.0');
```

src/wooden-tree/index.js
```diff
@@ -0,0 +1,7 @@
+import { Tree } from 'react-wooden-tree';
+
+export { ActionTypes, callBack, Node, treeDataReducer } from 'react-wooden-tree';
+require('./index.scss');
```
Hyperkid123: you can use import here too `import './index.scss'`

Hyperkid123:
> @Hyperkid123 If I try to load font-awesome (added as dev-dep) it tells me that I need an appropriate loader for the file:
> 
> ```
> ERROR in ./node_modules/font-awesome/css/font-awesome.css 7:0
> Module parse failed: Unexpected character '@' (7:0)
> You may need an appropriate loader to handle this file type.
> | /* FONT PATH
> |  * -------------------------- */
> > @font-face {
> |   font-family: 'FontAwesome';
> |   src: url('../fonts/fontawesome-webfont.eot?v=4.7.0');
> ```

We might not have appropriate loader i will check that out. My question would be why we need to add new font as dev dependency? Font awesome is already part of PF-react. and icons should are used in other components.

brumik:
I am sorry, yes, I used it and it works correctly without any other modification.

src/wooden-tree/index.js
brumik: Why it is better? How important it is?

Hyperkid123: require and import are equivalent. Require is pre ES6. You should not mix them. We went with import in this repo.

Also you should call root(your index.scss) scss file `style.scss`. Once you run build it will automatically create `wooden-tree.css` file which you need to import in your app to have these styles. It will ignore anything else.

brumik:
Who do I need to contact to merge this? @Hyperkid123

src/wooden-tree/index.js
```diff
@@ -0,0 +1,7 @@
...
+
+class WoodenTreeView extends Tree {}
+export default WoodenTreeView;
```
skateman: Do you insist this name? Or can we just name it TreeView?

skateman: Ah, nevermind...

Hyperkid123:
@brumik @skateman  let me test it

src/wooden-tree/stories/WoodenTree.stories.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+import { Provider } from 'react-redux';
+import { storiesOf } from '@storybook/react';
+import { store, WoodenTreeExample } from './WoodenTreeExample';
+
+storiesOf('WoodenTree', module)
+  .add('WoodenTreeExample', () => <Provider store={store}><WoodenTreeExample /></Provider>);
```
Hyperkid123: I have a small nitpick here. Storybook should allow you to copy and paste the code to your project and give you a working component. Instead pf `<Provider store={store}><WoodenTreeExample /></Provider>` you should add whole markup inside `import { store, WoodenTreeExample } from './WoodenTreeExample';`

brumik: I am sorry, but I don't get what you mean? What markup?

Hyperkid123: The actual code from WoodenTreeExample. You should be able to just copy/paste the stuff from storybook and have working component.

Now you see in the storybook this:  `<Provider store={store}><WoodenTreeExample /></Provider>`

And you should be able to see basically all the code from WoodenTreeExample:
```JSX
// callback definitions etc. Even the data structure possibly
<ReduxTree
  hierarchicalCheck
  showCheckbox
  multiSelect={false}
  preventDeselect
  allowReselect
   checkboxFirst
   connectedNode={ConnectedNode}
   nodeIcon="fa fa-file-o"
   callbacks={{
      onDataChange: this.onDataChange,
       lazyLoad: this.lazyLoad,
      }}
      />
// whatever else you need.
```
`storiesOf(..).add(...)` must return component but whatever code you write inside of this will be visible on the storybook page. I would put at minimum the callback definitions so we know what arguments they take and what they should return. That is crucial information.

Hyperkid123:
@brumik ok should be fine just fix your snapshots. It was probably caused by higher version of PF just update snapshots and it should be fine.

brumik:
Locally it passes all tests and there are no snapshots to upgrade. The local version is the same as the repository version here. @Hyperkid123 I don't know what can be the problem.

Hyperkid123:
@brumik you have never version of dependencies that is fine just run `yarn test -u`

brumik:
@Hyperkid123 can you force Travis to re-run the second tests (which is not passing)?

Hyperkid123:
@brumik i think you will need to skip more snapshots.

Hyperkid123:
@brumik nice i will see what is going on with the snapshots once i have some window.

brumik:
Okay :) :+1:

src/wooden-tree/index.js
karelhala: Please do not mix require and import at the same file. Unless you really have to use `require` for instance for `requireContext`

PR #121, opened 2019-06-11
Add support for button link in generic table row

mzazrivec:
remote-region-button-after

https://bugzilla.redhat.com/show_bug.cgi?id=1678142


martinpovolny:
TODO: check if the generated `<button>` tag has type button so that if we ever use (I think we do not now) the textual summary in a `<form>`, the form would not get submitted once the button is pressed.

PR #123, opened 2019-06-12
Add react-select prefix for Tagging select

PanSpagetka:
Add react-select prefix for Tagging select, so now all div inside select component should have react-select prefix with `__` separator. For example: `react-select__value-container` or `react-select__control`


PanSpagetka:
ping @karelhala

PR #124, opened 2019-06-13
Icon picker

fhlavac:
@Hyperkid123 @martinpovolny @terezanovotna 

- added array.prototype.flat to setupTests.js
- created font-icon, icon-list and icon-picker components
- created tests for font-icon, icon-list, icon-picker components and icon-picker-helper
- added icon-picker to storybook

### DIFF
--- a/demo/demo-app.jsx
+++ b/demo/demo-app.jsx
@@ -1,41 +1,16 @@
 import * as ReactDOM from 'react-dom';
 import * as React from 'react';
-import { Form, Field } from 'react-final-form';
-import { Form as PfForm, Col, Row, Button, Grid } from 'patternfly-react';
-import { FinalFormSwitch } from '../src/forms';
+import { Grid } from 'patternfly-react';
+import { IconPicker } from '../src/icon-picker';
 
-const onSubmit = values => console.log('onSubmit: ', values);
+const IconTypes = [
+  { type: 'pf', title: 'Patternfly' },
+  { type: 'fa', title: 'Font Awesome' },
+];
 
 const wrapperComponent = () => (
   <Grid>
-    <h1>Select</h1>
-    <Form
-      onSubmit={onSubmit}
-      initialValues={{
-        switch: false,
-      }}
-      render={({ handleSubmit }) => (
-        <PfForm
-          horizontal
-          onSubmit={handleSubmit}
-        >
-          <Row>
-            <Col xs={12}>
-              <Field
-                name="switch"
-                id="switch"
-                render={({ input, meta }) => <FinalFormSwitch input={input} meta={meta} label="Switch component" />}
-              />
-            </Col>
-          </Row>
-          <Row>
-            <Col xs={12}>
-              <Button type="submit">Submit</Button>
-            </Col>
-          </Row>
-        </PfForm>
-      )}
-    />
+    <IconPicker iconTypes={IconTypes} />
   </Grid>
 );
 
--- a/scripts/setupTests.js
+++ b/scripts/setupTests.js
@@ -10,3 +10,13 @@ configure({ adapter: new Adapter() });
 const dom = new JSDOM();
 global.document = dom.window.document;
 global.d3 = d3;
+
+// mock flat
+if (!Array.prototype.flat) {
+  Object.defineProperty(Array.prototype, 'flat', { // eslint-disable-line no-extend-native
+    value(depth = 1) {
+      return this.reduce((flat, toFlatten) =>
+        flat.concat((Array.isArray(toFlatten) && (depth > 1)) ? toFlatten.flat(depth - 1) : toFlatten), []);
+    },
+  });
+}
--- a/src/icon-picker/font-icon.jsx
+++ b/src/icon-picker/font-icon.jsx
@@ -1,0 +1,10 @@
+import React, { memo } from 'react';
+import PropTypes from 'prop-types';
+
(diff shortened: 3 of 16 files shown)


fhlavac:

terezanovotna:
Can we add more context? what is this supposed to do? where is it located? what is the goal of this picker? before and after and link to storybook?

Hyperkid123:
@terezanovotna yep sorry about lack of info.

So this is a component where user can pick an icon and that is stored somewhere. For instance when user is creating custom button he can select an icon from this icon picker. It is also being used in toolbars.

Icons are divided into tabs based on the font and unfortunately they don't have fixed size. Every one is a bit different. That is why the second tab is more narrow than the first. 

What we don't know is the spacing/sizing of the icons, color of selected icon (that red rectangle). Because this is a re-write from angular to React it was never properly reviewed by UX.

I will also make sure that we have online storybook ASAP.

Hyperkid123:
@terezanovotna storybook: http://miq-icon-picker.surge.sh/?selectedKind=Icon%20Picker&selectedStory=Icon%20Picker&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel

You should see just one empty button.

If not go to Icon Picker > Icon Picker tab in menu

martinpovolny:
The test failure seems relevant. Pls, fix it.

martinpovolny:
Looks great.

@skateman : how much work is replacing the current (angular) implementation with this one?

skateman:
@martinpovolny I don't think it's too much as we're able to call the foo_field_changed from reducers much easier than from angular. I'll take a look at it after I'm done with the remote console stuff.

fhlavac:
@skateman 

- Improved icons alignment: same width at all tabs, added vertical spacing
- Changed color of selected icon
- Added highlighting of selected icon cell instead of highlighting the icon itself
- Fixed test failure

There is something wrong with storybook but sometimes works: http://icon-picker.surge.sh/?selectedKind=Icon%20Picker&selectedStory=Icon%20Picker&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel

skateman:
@fhlavac I like it, can you please add the `font-fabulous` icons to the demo as well? Just to make sure if it works...

Hyperkid123:
@skateman we don't font-fabulous in react-ui-components

Hyperkid123:
@fhlavac can you update the screenshots in description please?

fhlavac:
@miq-bot remove_label wip

skateman:
@Hyperkid123 ok, can you guarantee me that it will work in ui-classic?

Hyperkid123:
We can link it and tests. But it will work as long as there is a stylesheet with `ff` prefix..

skateman:
Well, isn't it easier just to add it here as a vendor/dev dependency? It doesn't have to be bundled with the final js/css.

Hyperkid123:
@skateman Its not used anywhere. We will just link it and check it. Its no different from any other style sheet.

Hyperkid123:
@epwinchell by anywhere I ment anywhere in react-ui-components. I am aware that is used all over the place in manageiq.

FF was never par of react-ui-components. 
Icon picker will have FF icons if you tell the component to use it. You can specify which fonts/icons it will use in component props. I just don't want to bake it into react-ui-components.

skateman:
@Hyperkid123 if you want 1:1 symmetrical visual behavior with the icon picker in angular-ui-components, then IMO it should be visually as much accurate as possible. Also saying that it's not being used anywhere in the existing demo is not a reason to not promote our own stuff across projects. It would be used here ...

Hyperkid123:
@skateman i have no idea what is the problem here. This icon picker supports font-fabulous and all the fonts that the angular does, its  just not in the demo. Am i missing something? We just don't want another dependency here just because of storybook that does not make any sense. Icon picker was not made specifically for font-fabulous or for any other font.

1:1 symmetrical visual behavior does not require any specific font. It should look exactly the same for every font. And if it does not then we have a visual issue which should be fixed. I really don't get what is the problem here.

Hyperkid123:
@epwinchell i think i misunderstood. We will add it to the demo.

Hyperkid123:
> @Hyperkid123 if you want 1:1 symmetrical visual behavior with the icon picker in angular-ui-component

@skateman how is this relevant for different font-types? I though the icon picker should look the same for every font. And if it does not then we have a problem.

> Also saying that it's not being used anywhere in the existing demo is not a reason to not promote our own stuff across projects. It would be used here ...

Thanks to PF react-ui-components has now somewhere around 21mb of static assests that are being exported with the library. I want to avoid any other dependencies as much as possible. I think you should understand that. But i do agree with the promotion so we sill add it as a dev dependency so it won be bundled into the library itself.

Hyperkid123:
@skateman @epwinchell since this PR adds icon picker, once its merged we will create another PR which adds FF to the storybook later? I don't want to add stuff to @fhlavac which he was not supposed to do. What do you say about that?

skateman:
@Hyperkid123 is it possible to not export the font-fabulous with the package?

Hyperkid123:
@skateman yep as i said i will add it to dev dependencies and use it only in the storybook for this component. I just don't want add this tasks to @fhlavac.

skateman:
I'm okay with that, thanks!

PR #125, opened 2019-06-14
Add gettext string extraction

mzazrivec:
* add react-gettext-parser to extract i18n strings (plus config and friends)
* add missing gettext annotations
* add catalog `locale/react-ui-components.pot` (with extracted strings)

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,8 @@
     "build-storybook": "build-storybook -c scripts/storybook -o .out",
     "coveralls": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
     "prepublishOnly": "npm run build",
-    "release": "semantic-release"
+    "release": "semantic-release",
+    "gettext-extract": "yarn run react-gettext-parser --config ./scripts/react-gettext-parser-config.js 'src/**/{*.js,*.jsx}'"
   },
   "engines": {
     "node": ">= 6.0.0",
--- a/package.json
+++ b/package.json
@@ -99,6 +99,7 @@
     "jsdom": "^12.0.0",
     "mini-css-extract-plugin": "^0.4.0",
     "node-sass": "^4.9.2",
+    "react-gettext-parser": "^1.9.1",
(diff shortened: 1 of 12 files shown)


mzazrivec:
@himdel review please?

package.json
```diff
@@ -14,7 +14,8 @@
     "build-storybook": "build-storybook -c scripts/storybook -o .out",
     "coveralls": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
     "prepublishOnly": "npm run build",
-    "release": "semantic-release"
+    "release": "semantic-release",
+    "gettext-extract": "yarn run react-gettext-parser --config ./scripts/react-gettext-parser-config.js 'src/**/{*.js,*.jsx}'"
```
himdel: in ui-components, the task is called `gettext:extract` (colon vs dash), is that an intentional difference?

scripts/react-gettext-parser-config.js
```diff
@@ -0,0 +1,2 @@
+exports.output = 'locale/react-ui-components.pot';
+exports.funcArgumentsMap = { __: ['msgid'] };
```
himdel: ui-classic is also exposing `n__` , should we do the same here?

himdel:
The travis failures are only because you're using double quotes instead of single quotes.

(@Hyperkid123 I wonder if that eslint rule is an error intentionally?)

scripts/react-gettext-parser-config.js
mzazrivec: We could, although at the moment we don't use it anywhere in the repo. So I did not add it (yet).

Hyperkid123:
> The travis failures are only because you're using double quotes instead of single quotes.
> 
> (@Hyperkid123 I wonder if that eslint rule is an error intentionally?)

I think it is i did not configure this one. But quotes can be fixed by eslint itself automatically

mzazrivec:
@himdel CI is green now.

himdel:
Aaand merged :)

PR #127, opened 2019-06-20
Fix failing travis

martinpovolny:
This should fix failing travis on master.

ping @rvsia, @himdel 

https://travis-ci.org/ManageIQ/react-ui-components/builds/547878451?utm_medium=notification&utm_source=email


src/tagging/tests/__snapshots__/tag.modifier.test.js.snap
```diff
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Tagging modifier match snapshot 1`] = `
-<Fragment>
+<React.Fragment>
```
himdel: looks like this change is superfluous (and causes the new failure)

martinpovolny: That's that the snapshot update command did. Shall I revert it?

martinpovolny: It's probably about a version of something :-(

martinpovolny: Anyway did that.

PR #129, opened 2019-07-01
Edit TagGroup component to support onClick and displaying titles

hstastna:
**What:**
Edit `TagGroup` component to support `onClick` and displaying titles (if they are specified).
Add some data for the appropriate storybook.

**Why:**
Needed for https://github.com/ManageIQ/manageiq-ui-classic/pull/5675

**Related RFE:**
https://bugzilla.redhat.com/show_bug.cgi?id=1678124

Some screenshots at:
https://github.com/ManageIQ/manageiq-ui-classic/pull/5675

### DIFF
--- a/src/textual_summary/tag_group.jsx
+++ b/src/textual_summary/tag_group.jsx
@@ -7,29 +7,51 @@ import IconOrImage from './icon_or_image';
  */
 export default class TagGroup extends React.Component {
   /**
+   * Return onClick function if link is defined for an item/subitem
+   */
+  checkLinkItem = (item, e) => {
+    if (item.link && this.props.onClick) {
+      this.props.onClick(item, e);
+    }
+  }
+
+  /**
    * Render a simple row. Label, icon, value.
    */
   renderTagRowSimple = item => (
     <tr key={item.key}>
       <td className="label">{item.label}</td>
-      <td>
+      <td title={item.title} onClick={e => this.checkLinkItem(item, e)}>
         <IconOrImage icon={item.icon} image={item.image} title={item.title} />
+        {' '}
         {item.value}
       </td>
     </tr>
   );
 
   /**
-   * Render a list of values joined with "<b> | </b>"
+   * Render a list of values joined with "<b> | </b>", with or without label
    */
-  renderSubiteList = subitem => (
+  renderSubitemList = subitem => (
     <span>
-      {subitem.label}: {subitem.value.map((val, index) => (
+      &nbsp;
+      {subitem.label && `${subitem.label}: `}{Array.isArray(subitem.value) && subitem.value.map((val, index) => (
         <React.Fragment key={index}>
           {val}
           {(index < subitem.value.length - 1) && <b>&nbsp;|&nbsp;</b>}
         </React.Fragment>
-        ))}
+      ))}
+    </span>
+  );
+
+  /**
+   * Render a single value, with or without label
+   */
+  renderSubitem = subitem => (
+    <span>
+      &nbsp;
+      {subitem.label && `${subitem.label}: `}
+      {subitem.value}
     </span>
   );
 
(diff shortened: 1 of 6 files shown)


hstastna:
@miq-bot add_reviewer @rvsia

hstastna:

src/textual_summary/tag_group.jsx
```diff
@@ -46,11 +68,11 @@ export default class TagGroup extends React.Component {
...
         <tr key={index}>
           {(index === 0) && (
-          <td rowSpan={item.value.length} className="label">{item.label}</td>
+          <td rowSpan={item.value.length} className="label" title={item.title}>{item.label}</td>
             )}
-          <td>
+          <td title={subitem.title} onClick={e => this.checkLinkItem(subitem, e)}>
             <IconOrImage icon={subitem.icon} image={subitem.image} text={subitem.text} />
-            {this.renderSubiteList(subitem)}
+            {Array.isArray(subitem.value) ? this.renderSubiteList(subitem) : this.renderSubitem(subitem)}
```
Hyperkid123: Can you fix this typo `renderSubiteList`? I know it was here before :wink:

src/textual_summary/tag_group.jsx
```diff
@@ -6,30 +6,52 @@ import IconOrImage from './icon_or_image';
...
       </td>
     </tr>
   );
 
   /**
-   * Render a list of values joined with "<b> | </b>"
+   * Render a list of values joined with "<b> | </b>", with or without label
    */
-  renderSubiteList = subitem => (
+  renderSubitemList = subitem => (
```
himdel: :+1:

src/textual_summary/tag_group.jsx
```diff
@@ -6,30 +6,52 @@ import IconOrImage from './icon_or_image';
...
    * Render a simple row. Label, icon, value.
    */
   renderTagRowSimple = item => (
     <tr key={item.key}>
       <td className="label">{item.label}</td>
-      <td>
-        <IconOrImage icon={item.icon} image={item.image} title={item.title} />
-        {item.value}
+      <td title={item.title} onClick={e => this.checkLinkItem(item, e)}>
+        <IconOrImage icon={item.icon} image={item.image} title={item.title} /> {item.value}
```
himdel: keep the 2 lines as 2 lines please :)

src/textual_summary/tag_group.jsx
```diff
@@ -6,30 +6,52 @@ import IconOrImage from './icon_or_image';
...
+  );
+
+  /**
+   * Render a single value, with or without label
+   */
+  renderSubitem = subitem => (
+    <span>
+      &nbsp;
+      {subitem.label && `${subitem.label}: `}
+      <React.Fragment>
```
himdel: no need for a Fragment here, it was needed only because of the array map

himdel:
Overall, matches the other components which already do the same thing :+1: 

(I don't agree that onClick is ever the right way to go to a different screen, we should be using proper links, but that's a separate issue.)

src/textual_summary/tag_group.jsx
hstastna: Do you think it is better to add `&nbsp;` before `{item.value}` if I want to achieve a nice space between icon and the item name? Like this:
icon_space
Because I think we don't want this:
icon_close

Moving ``{item.value}` to the same line as `IconOrImage` was the simplest and working fix for this.

martinpovolny:
@himdel : all feedback properly adressed?

src/textual_summary/tag_group.jsx
himdel: Ah, sorry, react is not html...

I think the canonical way to do this (if I understand https://github.com/facebook/jsx/issues/19#issuecomment-57079949 right) is..

```jsx
<td title={item.title} onClick={e => this.checkLinkItem(item, e)}>
  <IconOrImage icon={item.icon} image={item.image} title={item.title} />
  {' '}
  {item.value}
</td>
```

himdel:
@martinpovolny No real issues from me, only https://github.com/ManageIQ/react-ui-components/pull/129#discussion_r301614590 but it's not really that important.

PR #131, opened 2019-07-03
One tag only feature

brumik:
One tag only mode allows only one tag from all categories by deleting all of the previously added tags.

* Added only one more reducer: `deleteAllAssignedTags`
* Updated tests with the new prop
* Fixed up the linter errors in the files which I touched (Mainly in the `tagging.stories.js` and `tagging.tests.js`)

@skateman @PanSpagetka 

Screenshot from 2019-07-04 11-52-24


brumik:
If this is OK can we merge? @PanSpagetka ?

skateman:
Gonna take a look too, gimme a minute...

skateman:
btw @brumik a screenshot would be nice in the description

skateman:
maybe it's a UX nitpick and @terezanovotna would say no, but I'd like to have both dropdowns "stuck" on the selected tag item if possible, having just one of them is confusing

brumik:
I don't understand your last comment, what you mean?

For the screenshot, the UI is the same that for the other tags (and there is a story in the storybook for it)

skateman:
Dropdown on the left after selecting a tag:
Screenshot from 2019-07-04 11-54-09

Upper one is stuck, the lower one is reset...IMO both should be the same value as it has been selected.

brumik:
I corrected that bug here: #132 which is forked from this branch

brumik:
With the styling:
Screenshot from 2019-07-04 12-59-30
Screenshot from 2019-07-04 12-59-43


Without it:
Screenshot from 2019-07-04 13-00-27
Screenshot from 2019-07-04 13-00-39


Note also the small X-es next to multiselect. If the UI implements it, it is working properly.

skateman:
Oh, I can live with that one!

brumik:
@Hyperkid123 Merge?

brumik:

src/tagging/reducers/reducers.js
```diff
@@ -1,5 +1,6 @@
 import * as actionsConstants from '../actions/actions';
-// state = state.assignedTags
+
+/* ================== SECTION: state = state.assignedTags ================== */
```
Hyperkid123: Can we rather use JS DOC notation instead of these comments?

brumik: Yes, but I didn't found how to make a section of code.

src/tagging/components/Tagging/Tagging.js
```diff
@@ -13,6 +13,13 @@ class Tagging extends React.Component {
       tagCategory: this.props.selectedTagCategory,
       tagValue: selectedTagValue,
     };
+
+    // Single Tag Only
+    if (this.props.options && this.props.options.onlySingleTag) {
+      this.props.onSingleTagValueChange(action);
+      return;
```
karelhala: No need to `return;` but rewrite it to use else if
```JS
if (this.props.options && this.props.options.onlySingleTag) {
  this.props.onSingleTagValueChange(action);
} else if (this.findSelectedTag(this.props.selectedTagCategory).singleValue) {
  this.props.onTagValueChange(action, this.props.options);
}
```

PR #132, opened 2019-07-03
Tagging multiselect box update

brumik:
Update the tag select value part so it displays correctly the selected values even if the category contains more than one value. This update should make the selection process a bit easier by showing the selected elements in the value select box and not showing them as an option.

@PanSpagetka @skateman


brumik:
The functionality is ready just the UX is not prepared for multi-select displaying in the select input field. It looks a bit strange. Also, there were bigger changes so if it is approved the tests should be updated.

@PanSpagetka @skateman

skateman:
Can you please include screenshots? I'm kinda bad with imagining something that you describe as "it looks strange" ... thanks.

brumik:
The functionality is working fine, but the UX is not prepared for multiselect:

**With the provided styling:**
Screenshot from 2019-07-04 12-59-30
Screenshot from 2019-07-04 12-59-43


**Without it (stock react-select styling):**
Screenshot from 2019-07-04 13-00-27
Screenshot from 2019-07-04 13-00-39


Note also the small X-es next to multiselect. If the UI implements it, it is working properly.

skateman:
I think we used to have a UX pattern for abbreviating stuff. You can set the [`ellipsis`](https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/) CSS rule and keept the whole text in the tooltip using the `title` attribute. I think you don't need to set this in the dropdown's body as that can be significantly wider, but the select itself should be consistent in size with the one above.

brumik:
Even implementing this would not solve it. The custom style forces all the options into one line, so with 4+ options even truncated it will overflow. Is it possible to change the custom styles?

brumik:
Waiting for #130 to merge - it fixes half of the problem

brumik:
@miq-bot add_reviewer @hyperkid123

src/tagging/components/InnerComponents/ValueSelector.js
```diff
@@ -6,51 +6,103 @@ import TaggingPropTypes from '../TaggingPropTypes';
...
-    this.props.onTagValueChange({
-      description: selectedOption.label,
-      id: selectedOption.value,
-    });
+  /**
+   * Maps the selected options for the select box input.
+   *
+   * @returns {id, label}[]
+   */
+  getActiveValues = () => {
```
PanSpagetka: Please try to rewrite it using `this.props.selectedOption.map`

package.json
```diff
@@ -130,7 +130,7 @@
     "react-dom": "^16.4.0",
     "react-final-form": "^4.1.0",
     "react-redux": "^5.0.6",
-    "react-select": "^2.4.2",
+    "react-select": "^3.0.4",
```
Hyperkid123: Did you made sure that nothing else broke? There are form components using this dependency.

src/tagging/components/InnerComponents/ValueSelector.js
```diff
@@ -6,51 +6,103 @@ import TaggingPropTypes from '../TaggingPropTypes';
...
+      this.props.onTagValueChange([]);
+      return;
+    }
+
+    // Multi select
+    if (Array.isArray(values)) {
+      const tagArray = [];
+
+      // Remap value to tag objects
+      for (let i = 0; i < values.length; i += 1) {
```
Hyperkid123: use `values.map`  instead.

src/tagging/components/InnerComponents/ValueSelector.js
```diff
@@ -6,51 +6,103 @@ import TaggingPropTypes from '../TaggingPropTypes';
...
+   * (happens when removing elements one by one).
+   * On reset button (multi select) sends the default empty array.
+   *
+   * @param values {label, value}[] Currently selected values.
+   */
+  handleChange = (values) => {
+    // Deleting all the multi selected items one by one
+    if (values === null) {
+      this.props.onTagValueChange([]);
+      return;
```
Hyperkid123: I would prefer seeing this `return this.props.onTagValueChange([]);` instead of empty return statement (i know it will have the same effect)

src/tagging/components/InnerComponents/ValueSelector.js
```diff
@@ -6,51 +6,103 @@ import TaggingPropTypes from '../TaggingPropTypes';
...
+      const tagArray = [];
+
+      // Remap value to tag objects
+      for (let i = 0; i < values.length; i += 1) {
+        tagArray.push({
+          description: values[i].label,
+          id: values[i].value,
+        });
+      }
+      this.props.onTagValueChange(tagArray);
```
Hyperkid123: ^^ `map` can be passed directly as a parameter.

src/tagging/components/Tagging/Tagging.js
```diff
@@ -16,31 +16,34 @@ class Tagging extends React.Component {
...
-    } else if (this.findSelectedTag(this.props.selectedTagCategory).singleValue) {
-      this.props.onTagValueChange(action, this.props.options);
     } else {
-      this.props.onTagMultiValueChange(action, this.props.options);
+      this.props.onTagValueChange(action, this.props.options);
     }
   };
 
-  onTagCategoryChange = selectedTagCategory =>
+  onTagCategoryChange = (selectedTagCategory) => {
```
Hyperkid123: there is no need for adding block around one liners

src/tagging/containers/tagging.js
```diff
@@ -1,8 +1,7 @@
 import { connect } from 'react-redux';
 import {
-  changeAssignedTag, deleteAssignedTag, deleteAllAssignedTags,
-  toggleTagCategoryChange, toggleTagValueChange, loadState,
-  addAssignedTag,
+  deleteAssignedTag, deleteAllAssignedTags, toggleTagCategoryChange,
```
Hyperkid123: Can you fix this to have one item per line please? (linter rule seems to be broken)

src/tagging/reducers/reducers.js
```diff
@@ -14,28 +14,27 @@ function deleteAssignedTag(state, actionTag) {
...
-  const filteredState = state.filter(tag => (tag.id !== actionTag.tagCategory.id));
-  const selectedItem = state.find(tag => (tag.id === actionTag.tagCategory.id)) || { values: [] };
-  const oldValues = selectedItem.values.filter(tagValue => (tagValue.id !== actionTag.tagValue.id));
-  return [...filteredState,
-    {
-      description: actionTag.tagCategory.description,
-      id: actionTag.tagCategory.id,
-      values: [...oldValues].concat([actionTag.tagValue]),
-    }];
+  if (actionTag.tagValue !== null && actionTag.tagValue.length > 0) {
```
Hyperkid123: Is it possible for `tagValue` to be `undefined?`

src/tagging/tests/__snapshots__/value.modifier.test.js.snap
```diff
@@ -28,12 +28,7 @@ exports[`TagCategory Component match snapshot 1`] = `
     <ValueSelector
       multiValue={false}
       onTagValueChange={[MockFunction]}
-      selectedOption={
```
Hyperkid123: Do you know what triggered this snapshot change? I can't see this test object being removed or changed anywhere.

Hyperkid123:
screenshot-github com-2019 07 24-09-35-29

This does not look right.

We have working styling for pf3 react-select here: https://github.com/data-driven-forms/react-forms/tree/master/packages/pf3-component-mapper/src/form-fields/select

src/tagging/tests/__snapshots__/value.modifier.test.js.snap
brumik: I did not update the test and I missed it. Thanks.

package.json
brumik: I double-checked: none of the select boxes had breaking changes and all of them was working (however there are some components which were broken before)

brumik:
> screenshot-github com-2019 07 24-09-35-29
> 
> This does not look right. The cross icon and the dropdown indicator should not wrap.
> 
> We have working styling for pf3 react-select here: https://github.com/data-driven-forms/react-forms/tree/master/packages/pf3-component-mapper/src/form-fields/select

For first the select box looks like this:
Screenshot from 2019-07-24 17-15-48

But I agree that it looks strange. However, I wasn't able to apply the styles on it. Is there anything more than the scss and the js custom styles?

Hyperkid123:
> But I agree that it looks strange. However, I wasn't able to apply the styles on it. Is there anything more than the scss and the js custom styles?

Let me check if this is broken on master as well.

Hyperkid123:
This is clearable multi select on current master.  There must be something in tagging which breaks it.

screenshot-localhost-6006-2019 07 25-09-18-28

src/forms/select-styles.js
```diff
@@ -138,6 +138,12 @@ const customStyles = {
   }),
   multiValue: (provided, { isDisabled }) => ({
     ...provided,
+    // Long multiple values truncating ~18chars
```
Hyperkid123: No styling changes should be necessary. No matter how it bad the component looks, it was created to match pf3 design and we should not change them.

brumik: So I just should remove the styling?

Hyperkid123: yup. Selec is styled as it should be. If you have some styling issues in the tagging component, you should create some local override. This should not be applied to the default styles.

brumik:
Screenshot from 2019-07-25 16-46-44

I created a fix in scss and it looks like this

Hyperkid123:
> Screenshot from 2019-07-25 16-46-44
> 
> I created a fix in a new scss and it looks like this

I got only one nitpick, the search input should be centered.

brumik:

src/tagging/components/InnerComponents/ValueSelector.js
```diff
@@ -4,53 +4,98 @@ import PropTypes from 'prop-types';
...
+   * @param values {label, value}[] Currently selected values.
+   */
+  handleChange = (values) => {
+    // Deleting all the multi selected items
+    if (values === null) {
+      return this.props.onTagValueChange([]);
+    }
+
+    // Multi select
+    if (Array.isArray(values)) {
```
karelhala: **Nitpick**: You can simplify this by adding inline `isArray` check and check if values exist. It looks more complicaed this way, so if you want you can change just the second part (check if values is array and if not true create new array with one item `values`)
```JS
const tagValues = values ? (Array.isArray(values) ? values : [values]) : [];
return this.props.onTagValueChange(tagValues.map(el => ({
  description: el.label,
  id: el.value,
}))
```

src/tagging/sass/multiSelectFix.scss
```diff
@@ -0,0 +1,6 @@
+.react-select__value-container.react-select__value-container--is-multi {
+  &.react-select__value-container--has-value {
+    padding-right: 22px !important;
```
karelhala: Isn't there a way how to make this rule more specific so we don't have to use `!important`? If not leave it be, but it would be nicer without it.

src/tagging/components/InnerComponents/ValueSelector.js
brumik: Nice, but I let the first check intact since linter doesn't like nested ternary operators.

brumik:
Yes, I just discovered, I am trying to change select box to the pf-select which should be well designed.

Hyperkid123:
> When you select a value in the multi select, you can see the value in input that is good. You can also see X button, but that button doesn't work. When you click on it behave same as any other place on select. Please either hide it or make it work.

Add a wrapper with some className and create style only for that case. Do not create global overrides for existing components.

src/tagging/components/InnerComponents/ValueSelector.js
karelhala: Yeah, sure. Just change the second one so it is simpler
```
return this.props.onTagValueChange((Array.isArray(values) ? values : [values]).map(el => ({
  description: el.label,
  id: el.value,
}))
```

PR #136, opened 2019-07-24
Set max width for tagging container

PanSpagetka:
In ManageIQ Tagging container is to wide, from buttons are out of the screen. In storybook it is not an issue, I think this is caused by merging PF4 styles, however I am not 100% sure.

Before:
Screenshot from 2019-07-24 10-27-22

After:
Screenshot from 2019-07-24 10-28-39


PanSpagetka:
ping @karelhala

PR #137, opened 2019-07-30
Change tags PropType from number to string or number

PanSpagetka:
Change tags PropType from number to string or number


PanSpagetka:
@karelhala

PR #138, opened 2019-08-05
tagging - reduce tagging.js bundle size by not including react-bootstrap

himdel:
Problem (ui-classic, courtesy of webpack-bundle-analyzer):

tagging bad

Running the same in react-ui-components:

uir-tagging bad

This is caused by including the whole of react-bootstrap in dist/tagging.js.
And we don't actually need it for a div with a role and a class, hence this PR :).

After:

uir-tagging good

And in storybook, it looks the same for both versions:

after


himdel:
Cc @PanSpagetka @Hyperkid123

Hyperkid123:
@himdel :+1: 

I think we should think about marking pf3/pf4 as externals and peer dependencies to further reduce the size.

PR #143, opened 2019-09-09
Use pf4 tagging styles

karelhala:
This PR fixes release and build on travis.


himdel:
If the problem is that relative imports should use a `./`, there's a couple more:

```
src/tagging-pf4/sass/style.scss
2:@import "tag-category";
3:@import "tag";
```

karelhala:
@himdel good catch. Fixed them as well.

PR #144, opened 2019-09-09
React tree higlights node when it is dirty

brumik:
Added a modified store to the react tree wrapper which supports the highlighting of the nodes with changed (dirty) checkboxes. The highlight colour is [#39A5DC](https://github.com/jonmiles/bootstrap-treeview/pull/273).
**Before**
Screenshot from 2019-09-09 15-45-25

**After**
Screenshot from 2019-09-09 15-44-19

Fixes https://github.com/ManageIQ/manageiq-ui-classic/issues/6011

@skateman 
@karelhala 
@Hyperkid123

### DIFF
--- a/src/wooden-tree/defaultStore.js
+++ b/src/wooden-tree/defaultStore.js
@@ -1,0 +1,45 @@
+import Tree, { ActionTypes } from './';
+
+const nodeCheckedWithDirty = (propNode, value) => {
+  let node = { ...propNode };
+  if (!node.state.disabled) {
+    if (!Object.hasOwnProperty.call(node.state, 'defaultChecked')) {
+      node = {
+        ...node,
+        state: {
+          ...node.state,
+          defaultChecked: node.state.checked,
+        },
+      };
+    }
+    node = {
+      ...Tree.nodeChecked(node, value),
+      classes: value !== node.state.defaultChecked ? 'dirty' : undefined,
+    };
+  }
+  return node;
+};
+
+const actionMapper = {
+  [ActionTypes.EXPANDED]: (state, value, node) => Tree.nodeUpdater(state, Tree.nodeExpanded(node, value)),
+  [ActionTypes.DISABLED]: (state, value, node) => Tree.nodeUpdater(state, Tree.nodeDisabled(node, value)),
+  [ActionTypes.SELECTED]: (state, value, node) => Tree.nodeUpdater(state, Tree.nodeSelected(node, value)),
+  [ActionTypes.CHILD_NODES]: (state, value, node) => Tree.nodeUpdater(state, Tree.nodeChildren(node, value)),
+  [ActionTypes.LOADING]: (state, value, node) => Tree.nodeUpdater(state, Tree.nodeLoading(node, value)),
+  [ActionTypes.CHECKED]: (state, value, node) => Tree.nodeUpdater(state, nodeCheckedWithDirty(node, value)),
+  [ActionTypes.ADD_NODES]: (state, value) => Tree.addNodes(state, value),
+};
+
+export default (state = {}, action) => {
+  let node;
+  if (action.nodeId) {
+    node = Tree.nodeSelector(state, action.nodeId);
+    if (!node) {
+      return state;
+    }
+  }
+
+  return Object.hasOwnProperty.call(actionMapper, action.type) ?
+    actionMapper[action.type](state, action.value, node) :
+    state;
+};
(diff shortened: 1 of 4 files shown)


brumik:

skateman:
Can you do the same for the redux tree as well? Or is it included in this already?

src/wooden-tree/defaultStore.js
```diff
@@ -0,0 +1,52 @@
...
+
+const actionMapper = {
+  [ActionTypes.EXPANDED]: Tree.nodeExpanded,
+  [ActionTypes.DISABLED]: Tree.nodeDisabled,
+  [ActionTypes.SELECTED]: Tree.nodeSelected,
+  [ActionTypes.CHILD_NODES]: Tree.nodeChildren,
+  [ActionTypes.LOADING]: Tree.nodeLoading,
+};
+
+export default (state = null, action) => {
```
Hyperkid123: I think that initial state should be something else than null. Empty object probably (or some structure that will work).

src/wooden-tree/defaultStore.js
```diff
@@ -0,0 +1,52 @@
...
+  if (!state) {
+    return state;
+  }
+
+  let node = null;
```
Hyperkid123: no need for `= null`

src/wooden-tree/defaultStore.js
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }
+
+  switch (action.type) {
+    case ActionTypes.EXPANDED:
+    case ActionTypes.SELECTED:
+    case ActionTypes.CHILD_NODES:
+    case ActionTypes.LOADING:
+    case ActionTypes.DISABLED:
+      if ({}.hasOwnProperty.call(actionMapper, action.type)) {
```
Hyperkid123: use `Object` instead of `{}`

src/wooden-tree/defaultStore.js
```diff
@@ -0,0 +1,52 @@
...
+        node = actionMapper[action.type](node, action.value);
+        return Tree.nodeUpdater(state, node);
+      }
+      return state;
+    case ActionTypes.CHECKED:
+      if (!node.state.disabled) {
+        if (!{}.hasOwnProperty.call(node.state, 'defaultChecked')) {
```
Hyperkid123: ^^

src/wooden-tree/defaultStore.js
```diff
@@ -0,0 +1,52 @@
...
+          node = { ...node, state: { ...node.state, defaultChecked: node.state.checked } };
+        }
+        node = {
+          ...Tree.nodeChecked(node, action.value),
+          classes: action.value !== node.state.defaultChecked ? 'dirty' : null,
```
Hyperkid123: Does the tree works with `null`. If not its better to use `undefined`

src/wooden-tree/defaultStore.js
```diff
@@ -0,0 +1,52 @@
+import Tree, { ActionTypes } from './';
+
+const actionMapper = {
+  [ActionTypes.EXPANDED]: Tree.nodeExpanded,
+  [ActionTypes.DISABLED]: Tree.nodeDisabled,
+  [ActionTypes.SELECTED]: Tree.nodeSelected,
+  [ActionTypes.CHILD_NODES]: Tree.nodeChildren,
+  [ActionTypes.LOADING]: Tree.nodeLoading,
```
karelhala: You should add `CHECKED` and `ADD_NODES` in here as well, so we don't have to keep this mapping on two places.

src/wooden-tree/defaultStore.js
```diff
@@ -0,0 +1,52 @@
...
+
+  let node = null;
+  if (action.nodeId) {
+    node = Tree.nodeSelector(state, action.nodeId);
+    if (!node) {
+      return state;
+    }
+  }
+
+  switch (action.type) {
```
karelhala: If you add `CHECKED` and `ADD_NODES ` in the mapping you can remove this switch and just use
```JS
return Object.hasOwnProperty.call(actionMapper, action.type) ? 
  actionMapper[action.type](state, action, node) :
  state
```

src/wooden-tree/defaultStore.js
```diff
@@ -0,0 +1,52 @@
+import Tree, { ActionTypes } from './';
+
+const actionMapper = {
+  [ActionTypes.EXPANDED]: Tree.nodeExpanded,
```
karelhala: You should update these callbacks with `Tree.nodeUpdater(state, node);` as well
```JS
[ActionTypes.EXPANDED]: (state, action, node) => (
  Tree.nodeUpdater(state, Tree.nodeExpanded(node, action.value))
)
```

PR #145, opened 2019-09-09
Fix dist/tagging.css

himdel:
This reverts a couple of deletions from #132 - cc @brumik 
Looks like all the styles contributing to `dist/tagging.css` were removed in c1a73f2, but looks like the intent was to only remove styles related to the old select, right?

And this mostly reverts #143, except for the linter fixes - cc @karelhala 
The idea is that this was right to fail, since the same issue caused ui-classic to fail as well.
So, reintroducing as a useful canary for next time :).

This means we should also be able to revert https://github.com/ManageIQ/manageiq-ui-classic/pull/6171 and just update to new react-ui-components, to get proper pf3 tagging styling back.

@miq-bot add_label wip
(this brings back all the styles, I *think* `tag-selector` should be the one to go, but want to make sure first)

### DIFF
--- a/src/tagging/index.js
+++ b/src/tagging/index.js
@@ -1,6 +1,7 @@
 import taggingApp from './reducers';
 import { TaggingConnected, TaggingWithButtonsConnected } from './containers/tagging';
 import TagView from './components/InnerComponents/TagView';
+import './sass/style.scss';
 
 export { taggingApp, TaggingConnected, TaggingWithButtonsConnected, TagView };
 export * from './components/Tagging/Tagging';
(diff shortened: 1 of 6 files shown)


himdel:
Removed most of tag-selector.scss, but `tag-icon` seems to have a meaning in TagSelector.js, not removing.

himdel:
Before (ui-classic with #132 and https://github.com/ManageIQ/manageiq-ui-classic/pull/6171):
bad

After (ui-classic with this PR and (ui PR TODO)):
good

himdel:
@miq-bot remove_label wip

martinpovolny:
Thanks all for the quick response!

djberg96:
:+1:

PR #156, opened 2019-10-23
Toolbar click close fix

martinpovolny:
Toolbar: pass onSelect and onKeyDown for drop-downs to properly close.

To be rebased after https://github.com/ManageIQ/react-ui-components/pull/155 is merged.

This makes also the keyboard work in the drop-downs. Meaning keyboard works everywhere except for the Kebab menu. Dealing with the Kebab menu is not part of this PR.


himdel:
Why all the `react/prop-types` exceptions? Wouldn't it be simpler to add those to propTypes, non-required?

PR #158, opened 2019-10-29
View button responsiveness fix

epwinchell:
The PR adds in the missing "toolbar-pf-action-right" class div, which makes the view buttons align left when in the responsive state.

Before (demo)
Screen Shot 2019-10-29 at 11 35 43 AM

After (demo)
Screen Shot 2019-10-29 at 11 29 29 AM

Before (ui classic)
Screen Shot 2019-10-29 at 11 34 25 AM

After (ui classic)
Screen Shot 2019-10-29 at 11 30 44 AM

### DIFF
--- a/src/toolbar/ToolbarView.jsx
+++ b/src/toolbar/ToolbarView.jsx
@@ -5,25 +5,27 @@ import { adjustColor } from './utility';
 const classNames = require('classnames');
 
 export const ToolbarView = ({ views, onClick }) => (
-  <div className="toolbar-pf-view-selector pull-right form-group">
-    {views.map(view => (
-      <button
-        key={view.id}
-        id={view.id}
-        name={view.name}
-        title={view.title}
-        className={classNames('btn btn-link', { active: view.selected })}
-        data-url={view.url}
-        data-url_parms={view.url_parms}
-        data-send_checked={view.send_checked ? 'true' : ''}
-        data-prompt={view.prompt}
-        data-popup={view.popup}
-        onClick={() => onClick(view)}
-      >
-        <i className={view.icon} style={{ color: adjustColor(view.color, view.enabled) }} />
-      </button>
-      ))
-    }
+  <div className="toolbar-pf-action-right">
+    <div className="form-group toolbar-pf-view-selector">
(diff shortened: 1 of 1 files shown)


epwinchell:
@miq-bot bug, cleanup

himdel:
@martinpovolny We really need to fix the jest bits, this also breaks ui travis :)

```
 FAIL  app/javascript/spec/toolbar/miq-toolbar.spec.jsx
  ● <MiqToolbar /> › renders ok for generic case
    expect(received).toMatchSnapshot()
    Snapshot name: `<MiqToolbar /> renders ok for generic case 1`
    - Snapshot
    + Received
    @@ -86,11 +86,15 @@
            <ToolbarView
              onClick={[Function]}
              views={Array []}
            >
              <div
    -           className="toolbar-pf-view-selector pull-right form-group"
    -         />
    +           className="toolbar-pf-action-right"
    +         >
    +           <div
    +             className="form-group toolbar-pf-view-selector"
    +           />
    +         </div>
            </ToolbarView>
          </div>
        </Toolbar>
      </MiqToolbar>
```

(regenerating the snapshot for now)

martinpovolny:
Yes, I suspected this would happen. I'll fix this the first thing tomorrow morning.

martinpovolny:
Oh you fixed it already.

martinpovolny:
The problem here is that I have to get the change merged to get the bot do a release.

After that, I can update the UI classic.

But at that point the test is already broken.

I think that what is needed here is remove the snapshot test from the UI classic (or move it here) and do a different type of test (w/o a snapshot) on UI classic side.

WDYT?

martinpovolny:
Btw, on top of thanking @himdel for quickly fixing this I also thank @Fryguy for quickly merging the fix ;-)

himdel:
Yup, either we should think about some way of getting the react-ui-components release (and ui-components) trigger a push to ui-classic (and ui-service) master, to use the new version right away,

or we can't do snapshot tests across repo boundaries.

Or both :).

I think there should be a way of telling jest "these are our components, expand them by default, but those are from elsewhere, stop at those".

PR #162, opened 2019-12-10
Toolbar: add 'kebabLimit' option.

martinpovolny:
Added option to configure the "kebabization" of the menu items.

 * kebabLimit === -1 means no compacting
 * kebabLimit === 0 means always compact into kebab
 * other values of kebabLimit give number of items to keep, before the rest is kebabized.

### TODO

 * ~~update specs~~
 * ~~add specs for the new behavior~~

### Testing

Running the `npm run server` gives you an example of behavior for different values of `kebabLimit`.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1683703

customb-2019-12-11_08-58

### Further changes:

 * update tests to make them pass (unrelated),
 * make `count` optional the same as `kebabLimit` (based on PR review feedback)


Hyperkid123:
@martinpovolny can you update the snapshot please?

demo/demo-app.jsx
```diff
@@ -10,12 +10,46 @@ const wrapperComponent = () => (
   <React.Fragment>
     <div className="toolbar-pf row">
       <Toolbar
+        count={1}
```
Hyperkid123: Can I ask what does the count does and why is it required?

martinpovolny: In manageiq the toolbar needs count of selected items in the GTL on the page.

Anyway: making it optional, adding default value.

src/toolbar/Toolbar.jsx
```diff
@@ -91,17 +107,23 @@ export const Toolbar = props => (
         .filter(toolbarGroupHasContent)
         .map((group, index) =>
           /* eslint react/no-array-index-key: "off" */
-          <ToolbarGroup key={index} onClick={props.onClick} group={collapseCustomGroups(group)} />)
+          <ToolbarGroup key={index} onClick={props.onClick} group={collapseCustomGroups(group, props.kebabLimit)} />)
```
Hyperkid123: I suppose we don't have anything better than index right?

martinpovolny: Not for the group. But the index is stable.

PR #163, opened 2020-01-02
Display blue text for a link, in SimpleTable component

hstastna:
**Issue:** https://github.com/ManageIQ/manageiq-ui-classic/issues/6572

This PR fixes the issue with the link to _Pod_ which did not look like a link, in the _Build Instances_ table in textual summary of selected _Container Build_.

**Before:**
before

**After:**
after1
after2


martinpovolny:
@epwinchell : what do you think about the color of the link? Blue is correct?

epwinchell:
> @epwinchell : what do you think about the color of the link? Blue is correct?

@martinpovolny Perfect

PR #166, opened 2020-02-13
Make searching through Tag categories and values work

hstastna:
**Related BZ:** https://bugzilla.redhat.com/show_bug.cgi?id=1797737

The related BZ says about a different bug (found in CFME 5.11/ivanchuk) but we need changes in this PR to be able to reproduce (and to fix) that bug also on the newest master, where we were not able to search for categories or values at all, while tagging selected items, because `isSearchable` was always set to `false`.

---

**Before:**
Selecting Tag Category:
category_before
Selecting Tag Value:
value_before

**After:**
category_after
If there is no _Category_ chosen yet, there is no reason to search through values:
value1_after
value2_after

### DIFF
--- a/src/tagging/components/InnerComponents/TagSelector.js
+++ b/src/tagging/components/InnerComponents/TagSelector.js
@@ -55,6 +55,7 @@ class TagSelector extends React.Component {
         id="tag_cat"
         // clearable={false}
         simpleValue={false}
+        searchable
       />
     );
   }
--- a/src/tagging/components/InnerComponents/ValueSelector.js
+++ b/src/tagging/components/InnerComponents/ValueSelector.js
@@ -48,6 +48,7 @@ class ValueSelector extends React.Component {
       multi={this.props.multiValue}
       simpleValue
       clearable
+      searchable={values.length > 0}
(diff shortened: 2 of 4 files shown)


hstastna:
@PanSpagetka @himdel @Hyperkid123 This PR is WIP because probably I will need to update the tests and because I am not 100% sure about these changes but you can review it anyway. Thank you in advance! :)

Hyperkid123:
@hstastna just update the jest snapshots please.

hstastna:
@Hyperkid123 I am not sure why it is still failing...

src/tagging/components/InnerComponents/TagSelector.js
```diff
@@ -55,6 +55,7 @@ class TagSelector extends React.Component {
         id="tag_cat"
         // clearable={false}
         simpleValue={false}
+        searchable={true}
```
PanSpagetka: Linter is failing here, you can just have `searchable` here without the `={true}` part.

hstastna: Ahh, great, I assumed that, but I was not sure. Thank you! :+1:

PR #167, opened 2020-02-17
Make repeatedly searching through Tag Categories and Values work

hstastna:
**Fixes:** https://bugzilla.redhat.com/show_bug.cgi?id=1797737
DEPENDS ON: https://github.com/ManageIQ/react-ui-components/pull/166 (merged)

There was an error regarding undefined `value[0]` after repeatedly searching through Tag Categories or Values. This was because `value` is an object with `label` and `value` keys, so using `value.label` instead of `value[0].label` works fine.

The bug was introduced in https://github.com/ManageIQ/react-ui-components/pull/109.

**Before:**
tag-before

**After:**
tag-after

### DIFF
--- a/src/forms/select-styles.js
+++ b/src/forms/select-styles.js
@@ -202,7 +202,7 @@ const customStyles = {
     const valueStub = {
       content: !isMulti // eslint-disable-line no-nested-ternary
         ? showPlaceholder ? `"${placeholder}"` // eslint-disable-line no-nested-ternary
-          : showValue ? `"${value[0].label}"` : '""' : '""',
+          : showValue ? `"${value.label}"` : '""' : '""',
       color: showPlaceholder ? '#999999' : '#363636',
       fontWeight: '600',
       minHeight: 20,


hstastna:
@miq-bot add_label bug

hstastna:
@Hyperkid123 @himdel

hstastna:
Just a note that regarding the BZ to 5.11, maybe some PRs will be needed to backport to ivanchuk branch.

mzazrivec:
We don't do branches in `react-ui-components`, right?

himdel:
Agreed, but we also dont use react-ui-components anywhere else.
So my plan is to create a manageiq-ui-classic ivanchuk/yes PR to move the component there,
instead of introducing a version branch.

himdel:
https://github.com/ManageIQ/react-ui-components/pull/168
https://github.com/ManageIQ/manageiq-ui-classic/pull/6691

hstastna:
**Note:**
I was testing changes in this PR also regarding selecting multiple values for one _Category_. Some categories allow us to choose more values at once.

PR #168, opened 2020-02-19
Move tagging to ui-classic

himdel:
Depends on https://github.com/ManageIQ/manageiq-ui-classic/pull/6691

We need to backport tagging separately from the rest,
we can either add react-ui-components release branches,
or move the component to ui-classic.

Doing the latter since that's the only place it's used.

One complication: `PfSelect` and the associated `select-styles.js` is used both by tagging and non-tagging code (report editor, forms), so..
* keeping it here,
* exporting just to make sure,
* creating a ui-classic copy in https://github.com/ManageIQ/manageiq-ui-classic/pull/6691 for now

Cc @hstastna 

@miq-bot add_label wip


himdel:
@miq-bot remove_label wip

hstastna:
https://github.com/ManageIQ/manageiq-ui-classic/pull/6691#issuecomment-590847814

PR #178, opened 2020-03-30
Fix react-wooden-tree icon size

brumik:
@miq-bot add_reviewer @mzazrivec 
@miq-bot add_label dependencies

**Before**
cc

**After**
Screenshot from 2020-04-07 14-16-58

**Reference**
Screenshot from 2020-04-07 17-27-09

**Note**: @skateman In the original tree all the icons are white not just the highlighted, but the new tree does not receive `iconColor` prop to paint them white.

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,7 +131,7 @@
     "react-final-form": "^4.1.0",
     "react-redux": "^5.0.6",
     "react-select": "^2.4.2",
-    "react-wooden-tree": "^2.1.3",
+    "react-wooden-tree": "^2.1.4",
     "redux": "^4.0.0",
     "redux-form-validators": "^2.7.0",
     "redux-mock-store": "^1.5.1"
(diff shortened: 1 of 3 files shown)


mzazrivec:
@brumik Would you happen to have before & after screenshots?

skateman:
@brumik there should be a little padding around the icon, see the current implementation for reference.

brumik:
@skateman The padding is used to between the icon and the text, I explicitly not colouring it. The colour is not on the icon btw but it is the "background colour". How important is to have that padding? I rather would not add another wrapper around the icon. I think it is not that bad :thinking:

skateman:
It should look as the original one :wink:

brumik:
@epwinchell

epwinchell:
@miq-bot add_reviewer @epwinchell

skateman:
https://github.com/ManageIQ/manageiq-ui-classic/pull/6958

PR #179, opened 2020-04-28
Delete the snapshot form and the unnecessary form components

skateman:
After https://github.com/ManageIQ/manageiq-ui-classic/pull/7004 we no longer need the snapshot form in this repo. This means that we can get rid of the final-form components and the explicit dependency on this library which blocks the DDF2 upgrade.

@miq-bot assign @himdel 
@miq-bot add_label cleanup, pending core
Parent issue: https://github.com/ManageIQ/manageiq-ui-classic/issues/6716
Depends on: https://github.com/ManageIQ/manageiq-ui-classic/pull/7004


himdel:
Looks like the final form bits were removed from all current providers so we can remove them :+1:  (created issue in nsxt, but that's not in manageiq yet).

himdel:
@skateman We'll also need to update guides - this is still documented as "the ddf" in https://github.com/ManageIQ/guides/blob/master/ui/react_forms.md

PR #181, opened 2020-09-03
Do not show hand pointer when links are not clickable

h-kataria:
Fixes https://github.com/ManageIQ/manageiq-ui-classic/issues/7216

before
before

after
after


himdel:
LGTM, but you may need to `jest -u` to update the snapshot tests. :)

PR #182, opened 2020-09-03
Disable an active twostate button

h-kataria:
Even tho button was active, user could re-click on it, this fix disables the button when it is active to eliminate the confusion.

Fixes https://github.com/ManageIQ/manageiq-ui-classic/issues/491

After fix
Screenshot from 2020-09-03 16-11-30

Screenshot from 2020-09-03 16-11-04


himdel:
LGTM :+1: :)

PR #184, opened 2020-11-09
Fix links color and A11y in table summary page

kavyanekkalapu:
Clickable items are not identifiable by color (usually they are in different color the UI) and you need to flyover on all fields to check which ones are clickable, links are not accessible with keyboard as well.

Before:
Screen Shot 2020-11-09 at 8 51 00 AM
 
After:
links are in different color and accessible with keyboard.
Screen Shot 2020-11-09 at 8 52 28 AM

Screen Shot 2020-11-09 at 8 52 42 AM

### DIFF
--- a/src/textual_summary/generic_table_row.jsx
+++ b/src/textual_summary/generic_table_row.jsx
@@ -33,12 +33,22 @@ export default function GenericTableRow(props) {
   const value = renderValue(item.value, props.onClick);
 
   return (
-    <tr onClick={e => props.onClick(item, e)} className={item.hoverClass} title={item.title}>
+    <tr className={item.hoverClass} title={item.title}>
       <td className="label">{item.label}</td>
+      {item.link &&
+      <td>
+        <a href={item.link} onClick={e => props.onClick(item, e)}>
+          <IconOrImage icon={item.icon} image={item.image} title={item.title} background={item.background} />
+          {item.button ? <Button bsSize="small">{value}</Button> : value}
+        </a>
+      </td>
+      }
+      {(item.link === undefined) &&
       <td>
         <IconOrImage icon={item.icon} image={item.image} title={item.title} background={item.background} />
         {item.button ? <Button bsSize="small">{value}</Button> : value}
       </td>
+      }
     </tr>
   );
 }
--- a/src/textual_summary/multilink_table.jsx
+++ b/src/textual_summary/multilink_table.jsx
@@ -9,10 +9,11 @@ export default function MultilinkTable(props) {
       title={sub.title}
       className="hover"
       style={{ cursor: 'pointer, hand' }}
-      onClick={e => onClick(sub, e)}
     >
-      <IconOrImage icon={sub.icon} image={sub.image} title={sub.title} />
-      {sub.value}
+      <a href={sub.link} onClick={e => onClick(sub, e)}>
+        <IconOrImage icon={sub.icon} image={sub.image} title={sub.title} />
+        {sub.value}
+      </a>
     </td>
   );
 
--- a/src/textual_summary/simple_table.jsx
+++ b/src/textual_summary/simple_table.jsx
@@ -28,7 +28,7 @@ export default function SimpleTable(props) {
       );
     } else if ((value != null) && (typeof value === 'object') && value.link) {
       // value with a link
-      return <td className="btn-link" onClick={e => onClick(value, e)} key={j}>{value.value}</td>;
+      return <td key={j} className="btn-link" ><a href={value.link} onClick={e => onClick(value, e)}>{value.value}</a></td>;
     }
     // simple value
     return <td className="no-hover" key={j}>{value}</td>;
--- a/src/textual_summary/table_list_view.jsx
+++ b/src/textual_summary/table_list_view.jsx
@@ -8,9 +8,11 @@ const simpleRow = (row, i, colOrder) => (
 );
 
 const clickableRow = (row, i, colOrder, rowLabel, onClick) => (
-  <tr key={i} onClick={e => onClick(row, e)}>
-    {colOrder.map((col, j) => <td key={j} title={rowLabel}>{`${row[col]}`}</td>)}
-  </tr>
+  <a href={row.link} onClick={e => onClick(row, e)}>
+    <tr key={i} >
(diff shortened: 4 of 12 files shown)


skateman:
@miq-bot add_reviewer @himdel

himdel:
* `src/textual_summary/generic_table_row.jsx`
  * IconOrImage+text inside a tr with onClick (fixed here)
* `src/textual_summary/multilink_table.jsx`
  * IconOrImage+text inside a td with onClick
* `src/textual_summary/table_list_view.jsx`
  * text inside tr with onClick
* `src/textual_summary/simple_table.jsx`
  * text inside td with onClick
* `src/textual_summary/tag_group.jsx`
  * icon+text inside td with onClick (multiple)

I *think* we would want all of those to work the same way?

himdel:
As far as the change itself goes, LGTM :+1:

But if it's really needed in all those places, maybe it makes sense to DRY it up into a td with optional link, text, icon, or component for the complex ones?

kavyanekkalapu:
@himdel thank you, will fix in remaining places too.

himdel:
2 issues:

* href + onClick is probably doing the wrong thing .. is there any need to keep `onClick` now?
 * if not, don't, links do not need an extra onclick
 * if yes, then the onclick code needs to know when to do the onclick and when to let the `<a href` do its thing

* all the code is now repeated twice, no reason for that, what about something like:

```javascript
const MaybeLink = ({item, children}) => item.link ? (<a href={item.link}>{children}</a>) : <>{children}</>;
```

kavyanekkalapu:
@himdel If it is anchor tag, we definitely need href, otherwise it is invalid and it will fail A11y too, refer https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md, solution is 
- making as button and styling it like a link, then we can remove href
- remove onclick function in a tag, i am not sure what is there in that onclick other than going to link

kavyanekkalapu:
@himdel Looks like onClick function has this code, so i can't remove onClick, only option is use button and styling it as ancor tag.
```  event.preventDefault();

  if (!item.link) {
    return;
  }

  if (item.external) {
    window.open(item.link, '_blank', 'noopener,noreferrer');
  } else if (item.explorer) {
    const tokenElement = document.querySelector("meta[name=csrf-token]");
    // FIXME: jQuery is necessary here as it communicates with the old world
    $.ajax({
      data: `authenticity_token=${encodeURIComponent(tokenElement ? tokenElement.getAttribute('content') : '')}`,
      dataType: 'script',
      type: 'post',
      url: item.link,
    });
  } else {
    window.DoNav(item.link);
  }

kavyanekkalapu:
@himdel Fixed href/Onclick redundancy  issue by replacing it with button and styling like an anchor tag

for second issue, i tried to write like this
```function MaybeLink(props) {
  if (props.item.link) {
    return <a href={props.item.link}>{props.children}</a>;
  }
  return (<div>{props.children}</div>);
}
but i got an error
<img width="1680" alt="Screen Shot 2020-11-12 at 1 04 31 PM" src="https://user-images.githubusercontent.com/37085529/98991162-75fe7880-24f9-11eb-91b2-e45d2098f3ef.png">
if i remove div, it is not working as well..

i am wondering if you can approve and merge pr if functionality is working well... after  that , if you have an idea about code optimization, you raise pr on top that, we are almost close to release, we need this code in kasparov branch. let me know your thoughts, thank you.

himdel:
> making as button and styling it like a link, then we can remove href

Button inside a table is not the right thing to do, and we'd lose the href, which was the original point of the exercise.

> if i remove div, it is not working as well..

In both cases, you're creating content inside a `<tr>` instead of `<td>`, hence the error

>  onClick function has this code, so i can't remove onClick

Ah, thanks, I think we can safely remove onclick, we just need 3 branches instead of 2:

  * not a link: `<>{children}</>`
  * non-explorer link: `<a href... target={item.external ? "_blank" : "_self"}>{children}</a>`
  * explorer: `<span onclick=..POST>{children}</span>`

Addressing in a separate PR :).

kavyanekkalapu:
> > making as button and styling it like a link, then we can remove href
> 
> Button inside a table is not the right thing to do, and we'd lose the href, which was the original point of the exercise.

Oh i don't know that, i remember seeing buttons inside table in MIQ like Services>Catalogs>Service Catalogs>All Services table, Could you explain why is wrong thing to do?
> 
> > if i remove div, it is not working as well..
> 
> In both cases, you're creating content inside a `<tr>` instead of `<td>`, hence the error
> 
> > onClick function has this code, so i can't remove onClick
> 
> Ah, thanks, I think we can safely remove onclick, we just need 3 branches instead of 2:
> 
> * not a link: `<>{children}</>`
> * non-explorer link: `<a href... target={item.external ? "_blank" : "_self"}>{children}</a>`
> * explorer: `<span onclick=..POST>{children}</span>`
> 
> Addressing in a separate PR :).

once your pr merges, then i make change to anchor tag?

h-kataria:
@himdel any updates on this PR, is this good to be merged?

himdel:
I do not consider this mergeable right now, sorry, @kavyanekkalapu if you can push the previous state (before the buttons), we can merge this, as the [issues](https://github.com/ManageIQ/react-ui-components/pull/184#issuecomment-725445662) there were only of code quality.

But .. a button is a clickable element with a background, a border and text, which can cause an action on the current screen.
A link is a clickable inline element, which can go to URL on the current screen or in a new tab.

There *is* a semantic difference, and what we're doing here are links, not buttons. Not to mention we're talking about a11y .. screenreaders do treat buttons and links differently. 
(Having to undo the button styling should be a cue that this is not the ideal course of action.)

---

Failing that, I'm about halfway with my PR, but no longer promising to get it ready today, too many distractions.

kavyanekkalapu:
@himdel Right now if i want put in previous state, that means anchor tag will have  onClick+href, is that OK with you? once your pr is merged , you can remove onclick, let me know your thoughts, thank you.

himdel:
@kavyanekkalapu Yes, I'm fine with that. It's still indicative of a code issue, but it does work :) 
Thanks! :)

kavyanekkalapu:
@himdel I updated the branch, can you check?

himdel:
Still working, LGTM, thanks! :)

PR #185, opened 2021-01-06
Introduced CSS Class to Prevent Table Labels Getting Cut Off

DavidResende0:
Related to: https://github.com/ManageIQ/manageiq-ui-classic/pull/7473

In conjunction with the above PR the `generic-row-label` class prevents the `word-break` property of row labels in textual summary tables for being overridden to `break-all` causing words in the label to get divided onto different lines.

Before:
image
After:
image

### DIFF
--- a/src/textual_summary/generic_table_row.jsx
+++ b/src/textual_summary/generic_table_row.jsx
@@ -34,7 +34,7 @@ export default function GenericTableRow(props) {
 
   return (
     <tr className={item.hoverClass} title={item.title}>
-      <td className="label">{item.label}</td>
+      <td className="label generic-row-label">{item.label}</td>
       {item.link &&
       <td>
         <a href={item.link} onClick={e => props.onClick(item, e)}>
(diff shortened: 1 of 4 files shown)


himdel:
This does ensure `word-break` instead of `break-all` for *some* textual summary tables.
LGTM, but .. is it intentionally just some of the tables?

himdel:
This https://github.com/ManageIQ/react-ui-components/pull/184#issuecomment-724066696 should help you find all the pieces.

himdel:
Should this be a class on the table, or just on the specific `td` or `th` inside?

DavidResende0:
Yeah wasn't sure what I was thinking the other day but it should just be on the specific `td` inside

DavidResende0:
> This does ensure `word-break` instead of `break-all` for _some_ textual summary tables.
> LGTM, but .. is it intentionally just some of the tables?

To answer your first question I think `generic_table_row.jsx` handles most tables, but the reason I didn't add the word-break elsewhere was because I didn't know everywhere the tables were being used (to test) and I tried to mess with the css as little as possible to prevent breaking something.

himdel:
Sure .. if you ever need to find those places, https://github.com/ManageIQ/react-ui-components/pull/184#issuecomment-724066696 should have examples for all of them :).

For now, merging as is :)

